PR #1, opened 2019-06-18
Ali&Leo: Flocx Provider

ljmcgann:
Hi Mainn, 

We created a separate folder called flocx_provider in the main directory that has all the code we wrote. 
We create two tables in the our db/ folder and also wrote functions to add data to tables in db/api.py. Most of the logic is implemented in root controller in api/controller/root.py. It starts from when the provider receives a post request from user under (POST_offers).

Let us know if you have any other questions.

Leo & Ali


flocx_provider/db/models.py
```diff
@@ -0,0 +1,42 @@
...
+class Offers(Base):
+    __tablename__ = 'provider_offers'
+    offer_id = Column(Integer, primary_key=True, autoincrement=True)
+    node_id = Column(Integer)
+    start_time = Column(String(50))
+    duration = Column(String(50))
+    end_time = Column(String(50))
+    price = Column(Integer)
+    ironic_config = Column(String(2000))
+    status = Column(Integer)
```
tzumainn: Why not string?

flocx_provider/db/models.py
```diff
@@ -0,0 +1,42 @@
...
+
+class Offers(Base):
+    __tablename__ = 'provider_offers'
+    offer_id = Column(Integer, primary_key=True, autoincrement=True)
+    node_id = Column(Integer)
+    start_time = Column(String(50))
+    duration = Column(String(50))
+    end_time = Column(String(50))
+    price = Column(Integer)
+    ironic_config = Column(String(2000))
```
tzumainn: You may want something like a dict here.

flocx_provider/db/models.py
```diff
@@ -0,0 +1,42 @@
...
+Base = declarative.declarative_base()
+
+class Offers(Base):
+    __tablename__ = 'provider_offers'
+    offer_id = Column(Integer, primary_key=True, autoincrement=True)
+    node_id = Column(Integer)
+    start_time = Column(String(50))
+    duration = Column(String(50))
+    end_time = Column(String(50))
+    price = Column(Integer)
```
tzumainn: Can price be a float?

flocx_provider/db/models.py
```diff
@@ -0,0 +1,42 @@
...
+    ironic_config = Column(String(2000))
+    status = Column(Integer)
+
+
+class ironic_config(Base):
+    __tablename__ = 'ironic_config'
+    node_id = Column(Integer, primary_key=True)
+    config = Column(String(50))
```
tzumainn: I don't think there's really a need to store an ironic config and duplicate ironic information; you can just grab it from ironic when needed.

PR #2, opened 2019-07-12
provider send an offer to marketplace

DanNiESh:
Implemented send_offer function in objects layer and test_send_offer code.
Provider can make a POST request to send offers by uuid to marketplace using keystoneauth session.
Still need to:
- replace hard-coded places with oslo config (keystone/marketplace endpoint)
- modify offer data model(from provider/marketplace side) to make the post request data consistent with each side.

### DIFF
--- a/esi_leap/conf/flocx_market.py
+++ b/esi_leap/conf/flocx_market.py
@@ -1,0 +1,61 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import copy
+
+from keystoneauth1 import loading
+from oslo_config import cfg
+
+
+opts = []
+
+flocx_market_group = cfg.OptGroup(
+    'flocx_market',
+    title='flocx_market option')
+
+
+def register_opts(conf):
+    conf.register_opts(opts, group=flocx_market_group)
+    loading.register_session_conf_options(conf, flocx_market_group)
+    loading.register_auth_conf_options(conf, flocx_market_group)
+    loading.register_adapter_conf_options(conf, flocx_market_group)
+    conf.set_default('valid_interfaces', ['internal', 'public'],
+                     group=flocx_market_group)
+    conf.set_default('service_type', 'marketplace', group=flocx_market_group)
+
+
+def list_opts():
+    def add_options(opts, opts_to_add):
+        for new_opt in opts_to_add:
+            for opt in opts:
+                if opt.name == new_opt.name:
+                    break
+            else:
+                opts.append(new_opt)
+
+    opts_copy = copy.deepcopy(opts)
+    opts_copy.insert(0, loading.get_auth_common_conf_options()[0])
+
+    plugins = ['password', 'v2password', 'v3password']
+    for name in plugins:
+        plugin = loading.get_plugin_loader(name)
+        add_options(opts_copy, loading.get_auth_plugin_conf_options(plugin))
+    add_options(opts_copy, loading.get_session_conf_options())
+
+    adapter_opts = loading.get_adapter_conf_options(
+        include_deprecated=False)
+    # adding defaults for valid interfaces
+    cfg.set_defaults(adapter_opts, service_type='marketplace',
+                     valid_interfaces=['internal', 'public'])
+    add_options(opts_copy, adapter_opts)
+    opts_copy.sort(key=lambda x: x.name)
+    return opts_copy
--- a/esi_leap/objects/base.py
+++ b/esi_leap/objects/base.py
@@ -9,7 +9,6 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
-
 from oslo_log import log
 from oslo_versionedobjects import base as object_base
 
--- a/esi_leap/objects/flocx_market_client.py
+++ b/esi_leap/objects/flocx_market_client.py
@@ -1,0 +1,65 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+# borrowed from Nova
+import esi_leap.conf
+from oslo_log import log as logging
+
+LOG = logging.getLogger(__name__)
+CONF = esi_leap.conf.CONF
+
+
+class FlocxMarketClient(object):
+    """Client class for flocx-market."""
+
+    def __init__(self, adapter=None):
+        self._adapter = adapter
+        self._client = self._create_client()
+
+    def _create_client(self):
+        """Create the HTTP session accessing the flocx-market service."""
+        client = self._adapter
+        client.additional_headers = {'accept': 'application/json'}
+        return client
+
+    def get(self, url, version=None, global_request_id=None):
+        headers = ({request_id.INBOUND_HEADER: global_request_id}
+                   if global_request_id else {})
+        return self._client.get(url, microversion=version, headers=headers)
+
+    def post(self, url, data, version=None, global_request_id=None):
+        headers = ({request_id.INBOUND_HEADER: global_request_id}
+                   if global_request_id else {})
+        return self._client.post(url, json=data, microversion=version,
+                                 headers=headers)
+
+    def put(self, url, data, version=None, global_request_id=None):
+        kwargs = {'microversion': version,
+                  'headers': {request_id.INBOUND_HEADER:
+                              global_request_id} if global_request_id else {}}
+        if data is not None:
+            kwargs['json'] = data
+        return self._client.put(url, **kwargs)
+
+    def delete(self, url, version=None, global_request_id=None):
+        headers = ({request_id.INBOUND_HEADER: global_request_id}
+                   if global_request_id else {})
+        return self._client.delete(url, microversion=version, headers=headers)
+
+    def send_offer(self, data):
+        url = CONF.flocx_market.endpoint_override + '/offer'
+        r = self.post(url, data)
+        if r.status_code != 201:
+            LOG.warning(
+                "Failed to send an offer to flocx-market. Got HTTP %s: %s",
+                r.status_code,
+                r.text)
+        return r.status_code
--- a/esi_leap/objects/offer.py
+++ b/esi_leap/objects/offer.py
@@ -10,18 +10,21 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-from oslo_versionedobjects import base as versioned_objects_base
 
 from esi_leap.common import statuses
 from esi_leap.db import api as dbapi
 from esi_leap.objects import base
 import esi_leap.objects.contract
 from esi_leap.objects import fields
+from esi_leap.objects import flocx_market_client
 from esi_leap.resource_objects import resource_object_factory as ro_factory
-import json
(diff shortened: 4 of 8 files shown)


esi_leap/objects/base.py
```diff
@@ -47,7 +47,11 @@ def _attr_as_dict(field):
...
-
-        return dict((k, getattr(self, k))
-                    for k in self.fields
-                    if self.obj_attr_is_set(k))
+        d = {}
+        for k in self.fields:
+            if self.obj_attr_is_set(k):
+                value = getattr(self, k)
+                if isinstance(value, (datetime.date, datetime.datetime)):
+                    value = value.isoformat()
```
tzumainn: I personally don't love this; what if someone wants to work with the date as a date?

esi_leap/objects/offer.py
```diff
@@ -58,6 +62,44 @@ def get_all_by_status(cls, context, status):
...
+    def send(cls, context, offer_uuid):
+        db_offer = cls.dbapi.offer_get(context, offer_uuid)
+        o = cls._from_db_object(context, cls(), db_offer)
+
+        auth = v3.Password(auth_url='http://localhost:5000/v3',
+                           username='flocx-provider',
+                           password='password',
+                           project_name='service',
+                           user_domain_id='default',
+                           project_domain_id='default')
```
tzumainn: I'm guessing you already know that these will eventually have to be updated to be oslo configs : )

esi_leap/objects/offer.py
```diff
@@ -58,6 +62,44 @@ def get_all_by_status(cls, context, status):
             context, status)
         return cls._from_db_object_list(context, db_offers)
 
+    @classmethod
+    def send(cls, context, offer_uuid):
+        db_offer = cls.dbapi.offer_get(context, offer_uuid)
+        o = cls._from_db_object(context, cls(), db_offer)
```
tzumainn: You can just use Offer.get here

esi_leap/objects/offer.py
```diff
@@ -58,6 +62,44 @@ def get_all_by_status(cls, context, status):
...
+        # remove unnecessary feilds
+        offer_dict.pop('created_at')
+        offer_dict.pop('updated_at')
+        offer_dict.pop('id')
+        offer_dict.pop('project_id')
+        offer_dict.pop('resource_type')
+        # fake fields
+        offer_dict['creator_id'] = 'Alice1992'
+        offer_dict['marketplace_date_created'] = '2016-07-16T19:20:30'
+        offer_dict['cost'] = 11.5
```
tzumainn: I'd advise creating another method for Offer called something like `to_marketplace_dict` which creates the dict that you're going to pass below. Then this code (along with the other changes) becomes something like:

```
o = Offer.get()
marketplace_offer_dict = o.to_marketplace_dict()
<create client, send the dict>
```

esi_leap/objects/base.py
DanNiESh: Yes, you're right. I'll remove it. I shouldn't put to_string method here. Date to string is only needed when converting into marketplace offer dictionary.

esi_leap/objects/offer.py
DanNiESh: Sure!

esi_leap/objects/offer.py
DanNiESh: I'll do that!

esi_leap/objects/offer.py
DanNiESh: Hi Mainn, I'm confused here: should I pass the marketplace or provider's username and password here in order to make a session?

tzumainn: In the long term the Marketplace may not be the same OpenStack instance as the provider, and may provide the provider with credentials to talk to the marketplace. So what we want is for the provider to specify whatever username/password/endpoint that the marketplace says to use.

DanNiESh:
> Hi! This is an excellent start, and something you should be excited to demo!
> 
> I've left some comments. In addition, I think we may want to instantiate a client object to the marketplace; this way we don't have to duplicate what we're doing where we're passing in the marketplace auth_url/username/password/etc every time we want to communicate with the marketplace.
> 
> Normally this is done by calling the service's actual client as in https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/resource_objects/ironic_node.py#L24. flocx-market doesn't (yet) have that though.
> 
> The alternative is something like: https://github.com/openstack/nova/blob/master/nova/scheduler/client/report.py#L184. I'd advise trying something like that.

Hi! I've looked into nova code on creating a client. If I got it correctly, there should be a FlocxMarketClient class with create_client, get/post/put/delete functions. (Also, I need to write test for it) Then, use this client object to send offers to marketplace. Should I make a separate PR for it?

tzumainn:
A separate PR probably makes sense, but if you wanted to combine it with this one, I'd be okay with that.

Note that the get/put/post functions in the client are called by *other* functions within the client, so I'd expect that your client would have a function such as 'create_offer` or something.

DanNiESh:
New changes:
- create to_marketplace_dict function
- use Offer.get() when fetching an offer to send to marketplace
- create marketplace client and use the client to send post request
Still cannot solve:
- keystoneauth1.loading auth/sess not searching for config file in default path

esi_leap/objects/flocx_market_client.py
```diff
@@ -0,0 +1,68 @@
...
+LOG = logging.getLogger(__name__)
+
+class FlocxMarketClient(object):	
+    """Client class for flocx-market."""
+    def __init__(self, adapter=None):
+        self._adapter = adapter
+        self._client = self._create_client()
+
+    def _create_client(self):
+        """Create the HTTP session accessing the placement service."""
```
tzumainn: "flocx-market service" : )

esi_leap/objects/offer.py
```diff
@@ -58,6 +66,29 @@ def get_all_by_status(cls, context, status):
             context, status)
         return cls._from_db_object_list(context, db_offers)
 
+    @classmethod
```
tzumainn: This doesn't need to be a class method; you can imagine it being called as

```
o = Offer.create(context, <values>)
o.send_to_flocx_market(context)
```

Also maybe call it `send_to_flocx_market` : )

esi_leap/tests/objects/test_offer.py
```diff
@@ -146,3 +146,16 @@ def test_save(self):
...
             self.assertEqual(updated_at, o.updated_at)
+
+    def test_send(self):
+        offer_uuid = self.fake_offer['uuid']
+        with mock.patch.object(self.db_api, 'offer_get',
+                               autospec=True) as mock_offer_send:
+            mock_offer_send.return_value = self.fake_offer
+
+            res = offer.Offer.send(self.context, offer_uuid)
+
```
tzumainn: You probably also want to mock the call to `marketplace_client.create_offer(marketplace_offer_dict)`; after all, we're just testing that the `send` method calls `create_offer`. Remember, these are unit tests and not integration tests.

DanNiESh: I don't fully get it. marketplace_client.create_offer(marketplace_offer_dict) is a function in offer.py. I guess it should be separate from test stuff?

tzumainn: Yep - in a unit test of the `send_to_marketplace` function, we don't want to test code from elsewhere, such as `marketplace_client.create_offer`. We just want to verify that it's called as expected.

DanNiESh:
New changes:
1. change offer.send_to_flocx_market to instance method
2. mock the call to flocx_market_client.send_offer in test_offer.test_send_to_flocx
3. register flocx_market group in oslo configuration and replace hardcoded keystone.auth1 auth/sess with flocx_market config
4. an offer will be sent to marketplace as soon as it's created

DanNiESh:
should this test (https://github.com/CCI-MOC/esi-leap/blob/52c2a9e51d9541248826daf2b267e18488a389f7/esi_leap/tests/api/controllers/v1/test_offer.py#L22) mock the call to objects.offer.create? It cannot pass because lack of keystone authentication.

tzumainn:
> should this test (
> 
> https://github.com/CCI-MOC/esi-leap/blob/52c2a9e51d9541248826daf2b267e18488a389f7/esi_leap/tests/api/controllers/v1/test_offer.py#L22
> 
> ) mock the call to objects.offer.create? It cannot pass because lack of keystone authentication.

Whoops - yep! Good catch.

esi_leap/conf/flocx_market.py
```diff
@@ -0,0 +1,63 @@
...
+
+import copy
+
+from keystoneauth1 import loading
+from oslo_config import cfg
+
+
+opts = [
+    cfg.StrOpt('api_endpoint', 
+        default='http://localhost:8081'), ]
```
tzumainn: We probably don't want a default here at all.

esi_leap/objects/offer.py
```diff
@@ -88,3 +113,25 @@ def expire(self, context=None):
...
+        offer_dict.pop('updated_at')
+        offer_dict.pop('id')
+        offer_dict.pop('project_id')
+        offer_dict.pop('resource_type')
+        # fake fields
+        offer_dict['creator_id'] = 'Alice1992'
+        offer_dict['marketplace_date_created'] = '2016-07-16T19:20:30'
+        offer_dict['cost'] = 11.5
+
+        return offer_dict
```
tzumainn: We probably want a newline here

esi_leap/objects/offer.py
```diff
@@ -88,3 +113,25 @@ def expire(self, context=None):
...
+        offer_dict['server_id'] = offer_dict.pop('resource_uuid')
+        offer_dict['provider_id'] = offer_dict.pop('uuid')
+        # remove unnecessary feilds
+        offer_dict.pop('created_at')
+        offer_dict.pop('updated_at')
+        offer_dict.pop('id')
+        offer_dict.pop('project_id')
+        offer_dict.pop('resource_type')
+        # fake fields
+        offer_dict['creator_id'] = 'Alice1992'
```
tzumainn: Set this to the `self.project_id`

esi_leap/objects/offer.py
```diff
@@ -88,3 +113,25 @@ def expire(self, context=None):
...
+        offer_dict['marketplace_date_created'] = '2016-07-16T19:20:30'
```
tzumainn: Don't pass this value in; in theory flocx_market should default this value

tzumainn:
Tests work great! There's a few more minor comments, about removing test values.

In the longer run, we'll want to:

* allow the marketplace offer creation to fail without failing the create method
* set the status of the offer to something different once the marketplace offer is successfully created
* have a periodic job that looks for offers who have failed to create a marketplace offer, and retries them
* probably add a marketplace_offer_id field

And

* pass ironic node information

But this is an excellent step!

esi_leap/objects/offer.py
DanNiESh: @fvukelic I think marketplace should have a default 'marketplace_date_created' value so that provider doesn't need to send this value to flocx_market. Right now I just fake this value in order to avoid 500 error in flocx_market when sending a post request.

tzumainn: Ah - maybe use the current date here for now then.

DanNiESh: Besides, the offer model in provider doesn't have a cost field, so I fake a cost value and pass it to marketplace.

DanNiESh:
Fixed:
1. remove default api_endpoint value in conf/flocx_market.py
2. modify test offer values

esi_leap/objects/offer.py
```diff
@@ -88,3 +111,24 @@ def expire(self, context=None):
...
+        offer_dict['provider_id'] = offer_dict.pop('uuid')
+        offer_dict['creator_id'] = offer_dict.pop('project_id')
+
+        # remove unnecessary feilds
+        offer_dict.pop('created_at')
+        offer_dict.pop('updated_at')
+        offer_dict.pop('id')
+        offer_dict.pop('resource_type')
+        # fake fields
+        offer_dict['marketplace_date_created'] = '2016-07-16T19:20:30'
```
tzumainn: Can you set this to something like `datetime.datetime.utcnow()`?

DanNiESh: fixed

tzumainn:
Nice work! Do users now have to specify endpoint_override instead of api_endpoint? Either way, it tests successfully and works well.

DanNiESh:
> Nice work! Do users now have to specify endpoint_override instead of api_endpoint? Either way, it tests successfully and works well.



> Nice work! Do users now have to specify endpoint_override instead of api_endpoint? Either way, it tests successfully and works well.

Yes. It should be set in the config file. The comments for endpoint_override in config file says "Always use this endpoint URL for requests for this client (flocx-market)"

tzumainn:
If you have a moment, could you submit a followup PR updating the README to explain what settings to set to connect to flocx_market?

DanNiESh:
> If you have a moment, could you submit a followup PR updating the README to explain what settings to set to connect to flocx_market?

No problem!

PR #4, opened 2019-07-19
add flocx-market documentation

DanNiESh:


tzumainn:
Thanks! I think we want this in the flocx-market repository though, and only for the flocx-market service (and not the provider).

DanNiESh:
Okay. I'm rewriting the readme file in flocx-market.

PR #5, opened 2019-07-22
send node data in an offer

DanNiESh:
Implement adding dummy node data in the offer to marketplace.
Adding a new dummy node as follows:
cat <<EOF > /tmp/nodes/1718
{
	"project_owner_id": "project id of dummy node owner",
	"server_config": {
		"new attribute XYZ": "This is just a sample list of free-form attributes used for describing a server.",
		"cpu_type": "Intel Xeon",
		"cores": 16,
		"ram_gb": 512,
		"storage_type": "samsung SSD",
		"storage_size_gb": 204
	}
}
EOF


esi_leap/resource_objects/dummy_node.py
```diff
@@ -35,6 +35,11 @@ def get_project_id(self):
             node_dict = json.load(node_file)
         return node_dict.get("project_id", None)
 
+    def get_node_config(self):
+    	with open(self._path) as node_file:
+            node_dict = json.load(node_file)
+        return node_dict.get("server_config", None)
+
```
tzumainn: Let's add this method to test_node.py and ironic_node.py as well. For the former it can return an empty dict; for the latter, it can also return an empty dict along with a comment mentioning that it'l lbe updated later.

DanNiESh:
- Updated README.md

- Added get_node_config methods in test_node.py and ironic_node.py

tzumainn:
Works great - thanks!

PR #6, opened 2019-07-25
provider retrieves contracts from marketplace and creates contracts

DanNiESh:
1. add a new periodic job: checking for new contracts in marketplace (haven't test yet)
2. flocx_market_client can send a get request to marketplace to retrieve all contracts (request with get_all_contracts REST api for now)
3. the objects.contract has a method retrieve_from_flocx_market that create a flocx_market_client and get the contracts list
4. the split method takes a flocx-market contract dict argument and split it into several small contracts with one offer
5. pass the test for objects.contract.retrieve_from_flocx_market and objects.contract.split

### DIFF
--- a/esi_leap/common/flocx_market.py
+++ b/esi_leap/common/flocx_market.py
@@ -1,0 +1,121 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from esi_leap.common import statuses
+from esi_leap.objects import flocx_market_client
+from esi_leap.objects import offer
+import json
+from keystoneauth1 import adapter
+from keystoneauth1 import loading as ks_loading
+from oslo_config import cfg
+from oslo_log import log as logging
+
+CONF = cfg.CONF
+LOG = logging.getLogger(__name__)
+
+
+def get_flocx_market_client():
+    auth_plugin = ks_loading.load_auth_from_conf_options(
+        CONF, 'flocx_market')
+    sess = ks_loading.load_session_from_conf_options(CONF, 'flocx_market',
+                                                     auth=auth_plugin)
+    adpt = adapter.Adapter(
+        session=sess,
+        service_type='marketplace',
+        interface='public')
+    return flocx_market_client.FlocxMarketClient(adpt)
+
+
+def retrieve_from_flocx_market(context):
+    marketplace_client = get_flocx_market_client()
+    contract_list = get_contracts(context, marketplace_client)
+    return contract_list
+
+
+def get_contracts(context, marketplace_client):
+    contracts = []
+    contract_offer_list = get_ocr_by_status(marketplace_client)
+    if len(contract_offer_list) == 0:
+        return []
+    for contract_offer in contract_offer_list:
+        contract_data = {}
+        contract_id = contract_offer["contract_id"]
+        marketplace_offer_id = contract_offer["marketplace_offer_id"]
+        ocr_id = contract_offer["offer_contract_relationship_id"]
+        contract_data["marketplace_offer_contract_relationship_id"] = ocr_id
+        c = get_contract_by_id(contract_id, marketplace_client)
+        contract_data["start_date"] = c["start_time"]
+        contract_data["end_date"] = c["end_time"]
+        contract_data["status"] = statuses.OPEN
+        bid = get_bid_by_id(c["bid_id"], marketplace_client)
+        contract_data["project_id"] = bid["project_id"]
+        offer_marketplace = get_offer_by_id(
+            marketplace_offer_id, marketplace_client)
+        provider_offer_id = offer_marketplace["provider_offer_id"]
+        offer_provider = offer.Offer.get(context, provider_offer_id).to_dict()
+        contract_data["offer_uuid"] = provider_offer_id
+        contract_data["properties"] = offer_provider["properties"]
+        contracts.append(contract_data)
+    return contracts
+
+
+def get_ocr_by_status(marketplace_client):
+    url = CONF.flocx_market.endpoint_override + \
+        '/offer_contract_relationship?status=unretrieved'
+    r = marketplace_client.get(url)
+    if r.status_code != 200:
+        LOG.warning(
+            "Failed to retrieve contracts from flocx-market. Got HTTP %s: %s",
+            r.status_code,
+            r.text)
+        return []
+    if r.content is not None:
+        return json.loads(r.content)
+    else:
+        return []
+
+
+def get_contract_by_id(contract_id, marketplace_client):
+    url = CONF.flocx_market.endpoint_override + '/contract/' + contract_id
+    res = marketplace_client.get(url)
+    return json.loads(res.content)
+
+
+def get_bid_by_id(bid_id, marketplace_client):
+    url = CONF.flocx_market.endpoint_override + '/bid/' + bid_id
+    res = marketplace_client.get(url)
+    return json.loads(res.content)
+
+
+def get_offer_by_id(offer_id, marketplace_client):
+    url = CONF.flocx_market.endpoint_override + '/offer/' + offer_id
+    res = marketplace_client.get(url)
+    return json.loads(res.content)
+
+
+def update_contract(o_c_r_id, marketplace_client):
+    url = CONF.flocx_market.endpoint_override + \
+        '/offer_contract_relationship/' + o_c_r_id
+    data = {"status": "retrieved"}
+    r = marketplace_client.put(url, data)
+    if r.status_code != 200:
+        LOG.warning(
+            "Failed to update a contract to flocx-market. Got HTTP %s: %s",
+            r.status_code,
+            r.text)
+    return r.status_code
+
+
+def update_flocx_market_contract(o_c_r_id):
+    marketplace_client = get_flocx_market_client()
+    res_status_code = update_contract(o_c_r_id, marketplace_client)
+    return res_status_code
--- a/esi_leap/db/sqlalchemy/api.py
+++ b/esi_leap/db/sqlalchemy/api.py
@@ -253,7 +253,6 @@ def contract_get_all_by_status(context, status):
 def contract_create(context, values):
     contract_ref = models.Contract()
     values['uuid'] = uuidutils.generate_uuid()
-    values['project_id'] = context.project_id
     contract_ref.update(values)
     contract_ref.save(get_session())
     return contract_ref
--- a/esi_leap/db/sqlalchemy/models.py
+++ b/esi_leap/db/sqlalchemy/models.py
@@ -70,6 +70,8 @@ class Contract(Base):
     id = Column(Integer, primary_key=True, nullable=False, autoincrement=True)
     uuid = Column(String(36), nullable=False, unique=True)
     project_id = Column(String(255), nullable=False)
+    marketplace_offer_contract_relationship_id = Column(String(36),
+                                                        nullable=False)
     start_date = Column(DateTime)
     end_date = Column(DateTime)
     status = Column(String(15), nullable=False, default=statuses.OPEN)
@@ -77,7 +79,6 @@ class Contract(Base):
     offer_uuid = Column(String(36),
                         ForeignKey('offers.uuid'),
                         nullable=False)
-
     offer = orm.relationship(
         Offer,
         backref=orm.backref('offers'),
--- a/esi_leap/manager/service.py
+++ b/esi_leap/manager/service.py
@@ -10,21 +10,21 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+
+from esi_leap.common import flocx_market
+from esi_leap.common import statuses
+import esi_leap.conf
+from esi_leap.manager import utils
+from esi_leap.objects import contract
+from esi_leap.objects import offer
 from oslo_context import context as ctx
 from oslo_log import log as logging
 import oslo_messaging as messaging
 from oslo_service import service
 from oslo_utils import timeutils
 
-from esi_leap.common import statuses
-import esi_leap.conf
-from esi_leap.manager import utils
-from esi_leap.objects import contract
-from esi_leap.objects import offer
-
-
 CONF = esi_leap.conf.CONF
-EVENT_INTERVAL = 10
+EVENT_INTERVAL = 60
 LOG = logging.getLogger(__name__)
 
 
@@ -49,11 +49,13 @@ class ManagerService(service.Service):
         LOG.info("Starting esi-leap manager RPC server")
         self.tg.add_thread(self._server.start)
         LOG.info("Starting _fulfill_contracts periodic job")
-        self.tg.add_timer(300, self._fulfill_contracts)
+        self.tg.add_timer(EVENT_INTERVAL, self._fulfill_contracts)
         LOG.info("Starting _expire_contracts periodic job")
-        self.tg.add_timer(60, self._expire_contracts)
+        self.tg.add_timer(EVENT_INTERVAL, self._expire_contracts)
         LOG.info("Starting _expire_offers periodic job")
-        self.tg.add_timer(60, self._expire_offers)
+        self.tg.add_timer(EVENT_INTERVAL, self._expire_offers)
+        LOG.info("Starting _retrieve_contracts periodic job")
+        self.tg.add_timer(EVENT_INTERVAL, self._retrieve_contracts)
 
     def stop(self):
         super(ManagerService, self).stop()
@@ -69,6 +71,21 @@ class ManagerService(service.Service):
                c.start_date <= timeutils.utcnow():
                 LOG.info("Fulfilling contract %s", c.uuid)
                 c.fulfill(self._context)
+
+    def _retrieve_contracts(self):
+        LOG.info("Checking for new contracts in marketplace")
+        contracts = flocx_market.retrieve_from_flocx_market(self._context)
+        for c in contracts:
+            co = contract.Contract(self._context, **c)
+            co.create(self._context)
+            LOG.info("Creating contract %s", co.uuid)
+            o_c_r_id = c["marketplace_offer_contract_relationship_id"]
+            res_status_code = flocx_market.update_flocx_market_contract(
+                o_c_r_id)
+            if res_status_code == 200:
+                LOG.info(
+                    "Updating offer contract pair in marketplace for offer %s",
+                    o_c_r_id)
 
     def _expire_contracts(self):
         LOG.info("Checking for expiring contracts")
--- a/esi_leap/objects/contract.py
+++ b/esi_leap/objects/contract.py
@@ -10,13 +10,15 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-from oslo_versionedobjects import base as versioned_objects_base
-
 from esi_leap.common import statuses
 from esi_leap.db import api as dbapi
 from esi_leap.objects import base
 from esi_leap.objects import fields
 import esi_leap.objects.offer
+from oslo_config import cfg
+from oslo_versionedobjects import base as versioned_objects_base
+
+CONF = cfg.CONF
 
 
 @versioned_objects_base.VersionedObjectRegistry.register
@@ -32,6 +34,7 @@ class Contract(base.ESILEAPObject):
         'status': fields.StringField(),
         'properties': fields.FlexibleDictField(nullable=True),
         'offer_uuid': fields.UUIDField(),
+        'marketplace_offer_contract_relationship_id': fields.UUIDField()
     }
 
     @classmethod
--- a/esi_leap/objects/offer.py
+++ b/esi_leap/objects/offer.py
@@ -10,7 +10,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-
+import datetime
 from esi_leap.common import statuses
 from esi_leap.db import api as dbapi
 from esi_leap.objects import base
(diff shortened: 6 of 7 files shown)


esi_leap/objects/contract.py
```diff
@@ -99,3 +119,22 @@ def expire(self, context=None):
...
+            contract_dict['offer_uuid'] = o['uuid']
+            contract_dict['project_id'] = o['project_id']
+            contract_dict['start_date'] = c['start_time']
+            contract_dict['end_date'] = c['end_time']
+            contract_dict['status'] = c['status']
+            contract_dict['cost'] = o['properties'].get('floor_price', 0)
+            contract_dict['created_at'] = c['time_created']
+            contract_dict['updated_at'] = None
+            contracts.append(contract_dict)
+        return contracts
```
tzumainn: I think I'd put this method into something like `common/flocx_market.py`, just because it's not really a method that deals with provider contracts.

DanNiESh: I've moved this method to common/flocx_market.py. Since the get_contract api in marketplace for provider hasn't implement yet, provider cannot establish a real connection with marketplace. I'm writing test for pulling and pushing contracts to marketplace.

DanNiESh:
New update:
1. create a retrieve_from_marketplace job in manager service, successfully checks for new contracts in marketplace and updates the status of contracts in marketplace to be "retrieved"
2. in order to get the contract and offer data from marketplace, the provider makes multiple get requests: a. get all offer_contract_relationship (marketplace doesn't have get offer_contract_relationship by status for now) b. store the contract_id and offers_id in a dictionary (key: contract_id, value: [offer_id]) c. make get requests to according contracts and offers api to get the required data.
3. Update the offer_contract_relationship status to be "retrieved" for each contract offer pair in marketplace.

esi_leap/db/sqlalchemy/api.py
```diff
@@ -253,7 +253,8 @@ def contract_get_all_by_status(context, status):
 def contract_create(context, values):
     contract_ref = models.Contract()
     values['uuid'] = uuidutils.generate_uuid()
-    values['project_id'] = context.project_id
+    # modified: project_id for a contract should be the offer project_id??
+    # values['project_id'] = context.project_id
```
DanNiESh: I changed the code here, because I think the contract project_id should be the offer project_id. Correct me if I'm wrong.

esi_leap/objects/contract.py
```diff
@@ -62,6 +67,20 @@ def get_all_by_status(cls, context, status):
             context, status)
         return cls._from_db_object_list(context, db_contracts)
 
+    @classmethod
+    def retrieve_from_flocx_market(cls):
```
DanNiESh: Should this method belong to common/flocx-market? Since it's not relevant to contract in provider's database.

esi_leap/db/sqlalchemy/api.py
tzumainn: Not as currently written; the contract project_id should be set to the marketplace bid's project_id, as that is the value that is needed to give access to a consumer. It's possible that the column should be renamed long-term, but for now I think you're right to remove this default.

esi_leap/objects/contract.py
tzumainn: Yep, I think so!

esi_leap/objects/contract.py
```diff
@@ -99,3 +118,22 @@ def expire(self, context=None):
...
+    #         contract_dict['offer_uuid'] = o['uuid']
+    #         contract_dict['project_id'] = o['project_id']
+    #         contract_dict['start_date'] = c['start_time']
+    #         contract_dict['end_date'] = c['end_time']
+    #         contract_dict['status'] = c['status']
+    #         contract_dict['cost'] = o['properties'].get('floor_price', 0)
+    #         contract_dict['created_at'] = c['time_created']
+    #         contract_dict['updated_at'] = None
+    #         contracts.append(contract_dict)
+    #     return contracts
```
tzumainn: This can probably all just be removed then.

esi_leap/objects/offer.py
```diff
@@ -122,7 +121,9 @@ def to_marketplace_dict(self):
         offer_dict['end_time'] = offer_dict.pop('end_date').isoformat()
         offer_dict['cost'] = offer_dict['properties'].get('floor_price', 0)
         offer_dict['server_id'] = offer_dict.pop('resource_uuid')
-        offer_dict['provider_id'] = offer_dict.pop('uuid')
+        offer_dict['provider_offer_id'] = offer_dict.pop('uuid')
+        # does not have a provider id for now
+        offer_dict['provider_id'] = "0001"
```
tzumainn: Yeah, we got rid of this from the data model; you can just remove this.

esi_leap/objects/contract.py
tzumainn: I think common/flocx_market.py may have been accidentally left out of the PR?

esi_leap/objects/contract.py
```diff
@@ -99,3 +118,22 @@ def expire(self, context=None):
...
+    # @classmethod
+    # def split(cls, context, c):
+    #     # split contract with multiple offers into serval contracts
+    #     offers = c['offers']
+    #     contracts = []
+    #     for o_uuid in offers:
+    #         contract_dict = {}
+    #         o = esi_leap.objects.offer.Offer.get(context, o_uuid).to_dict()
+    #         contract_dict['offer_uuid'] = o['uuid']
+    #         contract_dict['project_id'] = o['project_id']
```
tzumainn: This will have to change - and may require an additional query to get the project_id from the bid.

esi_leap/objects/contract.py
DanNiESh: ops, I missed that file. Already added it

DanNiESh:
Update:
1. contract data model has a new field "offer_contract_relationship_id", which is used to make put request to offer_contract_relationship api.
2. for the issue with scoping contract_offer, context.is_admin need to be hardcoded to true in order for provider to make a put request to offer_contract_relationship. I set "context.is_admin = True before this line: https://github.com/CCI-MOC/flocx-market/blob/e4c9d88fd45c0a49f13e90e918bd154b86140e59/flocx_market/db/sqlalchemy/api.py#L304

esi_leap/objects/offer.py
DanNiESh: Yes! Already removed.

esi_leap/objects/contract.py
DanNiESh: I've moved this method to common/flocx_market.py. I will add tests for common/flocx_market right away.

esi_leap/db/sqlalchemy/models.py
```diff
@@ -77,7 +77,7 @@ class Contract(Base):
     offer_uuid = Column(String(36),
                         ForeignKey('offers.uuid'),
                         nullable=False)
-
+    marketplace_offer_contract_relationship_id = Column(String(36), nullable=False)
```
tzumainn: You'll want to move this up a few lines, maybe just below the project_id

tzumainn:
Thanks! Still some things to do, but I think this is getting there. One added suggestion: the functions in objects/flocx_market_client.py could probably just be moved into common/flocx_market.py.

DanNiESh:
> Thanks! Still some things to do, but I think this is getting there. One added suggestion: the functions in objects/flocx_market_client.py could probably just be moved into common/flocx_market.py.

should I move the entire get_contracts() and update_contract() functions into common/flocx_market.py or just extract the contract data processing part  into it?

tzumainn:
> > Thanks! Still some things to do, but I think this is getting there. One added suggestion: the functions in objects/flocx_market_client.py could probably just be moved into common/flocx_market.py.
> 
> should I move the entire get_contracts() and update_contract() functions into common/flocx_market.py or just extract the contract data processing part into it?

If it's updating the marketplace contracts, I'd move all of it.

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,53 @@
...
+def split(context, c):
+        # split contract with multiple offers into serval contracts
+    offers = c['offers']
+    o_c_rs = c['o_c_rs']
+    contracts = []
+    for i in range (len(offers)):
+        contract_dict = {}
+        o = offer.Offer.get(context, offers[i]).to_dict()
+        contract_dict['offer_uuid'] = str(o['uuid'])
+        contract_dict['project_id'] = str(o['project_id'])
```
tzumainn: This should actually be the bid's project_id. Looks like we need another query...

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,53 @@
...
+def retrieve_from_flocx_market():
+    auth_plugin = ks_loading.load_auth_from_conf_options(
+        CONF, 'flocx_market')
+    sess = ks_loading.load_session_from_conf_options(CONF, 'flocx_market',
+                                                         auth=auth_plugin)
+    adpt = adapter.Adapter(
+        session=sess,
+        service_type='marketplace',
+        interface='public')
+    marketplace_client = flocx_market_client.FlocxMarketClient(adpt)
```
tzumainn: You shouldn't have to repeatedly re-initialize the client; instead, do something like this: https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/resource_objects/ironic_node.py

esi_leap/objects/flocx_market_client.py
```diff
@@ -63,3 +64,64 @@ def send_offer(self, data):
...
+                    r_offer = self.get(url2)
+                    o = json.loads(r_offer.content)
+                    contract_data["offers"].append(str(o["provider_offer_id"]))
+                    contract_data["marketplace_offer_id"].append(str(o["marketplace_offer_id"]))
+                    contract_data["o_c_rs"].append(str(o_c_r[1]))
+                contracts.append(contract_data)
+            return contracts
+        else:
+            return []
+
```
tzumainn: You're trying to be efficient here with the hash table, to avoid repeated queries. However the code is a bit hard to read and debug, and in the long-term we'll have a single query to deal with all of this. I suggest doing this the straightforward way: after you retrieve all the ocrs, iterate over the ocrs, and for each ocr query the contract from the marketplace and then the bid from the marketplace.

esi_leap/objects/flocx_market_client.py
```diff
@@ -63,3 +64,64 @@ def send_offer(self, data):
...
+                    o_c_r_tuple = (str(c_o["marketplace_offer_id"]), str(c_o["offer_contract_relationship_id"]))
+                    contract_offer_table[str(c_o["contract_id"])] = [o_c_r_tuple]
+                    contract_offer_table[str(c_o["contract_id"])].append(o_c_r_tuple)
+
+            # retrieve contracts and offers data from flocx-market iteratively
+            for contract_id, o_c_rs in contract_offer_table.items():
+                contract_data = {}
+                url1 = CONF.flocx_market.endpoint_override + '/contract/' + contract_id
+                r_contract = self.get(url1)
+                c = json.loads(r_contract.content)
```
tzumainn: These few lines should probably go into their own `get_contract` function

esi_leap/db/sqlalchemy/models.py
DanNiESh: Done

esi_leap/common/flocx_market.py
DanNiESh: fixed. I created a new function get_flocx_market_client().

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,54 @@
...
+    o_c_rs = c['o_c_rs']
+    contracts = []
+    for i in range (len(offers)):
+        contract_dict = {}
+        o = offer.Offer.get(context, offers[i]).to_dict()
+        contract_dict['offer_uuid'] = str(o['uuid'])
+        contract_dict['project_id'] = str(o['project_id'])
+        contract_dict['start_date'] = c['start_time']
+        contract_dict['end_date'] = c['end_time']
+        contract_dict['status'] = c['status']
```
tzumainn: the provider actually wants this to be set to 'open'; I would `from esi_leap.common import statuses` and set this to `statuses.OPEN`

larsks: Oh hey, that's almost https://github.com/CCI-MOC/flocx-market/issues/121#issuecomment-518670095.

esi_leap/manager/service.py
```diff
@@ -54,6 +54,8 @@ def start(self):
         self.tg.add_timer(60, self._expire_contracts)
         LOG.info("Starting _expire_offers periodic job")
         self.tg.add_timer(60, self._expire_offers)
+        LOG.info("Starting _retrieve_contracts periodic job")
+        self.tg.add_timer(300, self._retrieve_contracts)
```
tzumainn: Can you set this and the other timers to EVENT_INTERVAL, and set EVENT_INTERVAL = 60?

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,54 @@
...
+
+def get_flocx_market_client():
+    auth_plugin = ks_loading.load_auth_from_conf_options(
+        CONF, 'flocx_market')
+    sess = ks_loading.load_session_from_conf_options(CONF, 'flocx_market',
+                                                         auth=auth_plugin)
+    adpt = adapter.Adapter(
+        session=sess,
+        service_type='marketplace',
+        interface='public')
```
larsks: We shouldn't be hardcoding the `interface` here, but that can be fixed in a future PR.

esi_leap/objects/flocx_market_client.py
```diff
@@ -63,3 +64,65 @@ def send_offer(self, data):
...
+
+    def get_contracts(self):
+        url = CONF.flocx_market.endpoint_override + \
+        '/offer_contract_relationship?status=unretrieved'
+        r = self.get(url)
+        if r.status_code != 200:
+            LOG.warning(
+                "Failed to retrieve contracts from flocx-market. Got HTTP %s: %s",
+                r.status_code,
+                r.text)
```
larsks: Shouldn't we `return` at this point, rather than falling through the rest of the function?

esi_leap/objects/contract.py
```diff
@@ -62,6 +68,20 @@ def get_all_by_status(cls, context, status):
             context, status)
         return cls._from_db_object_list(context, db_contracts)
 
+    # @classmethod
```
larsks: Can we just remove all of these commented-out lines?

esi_leap/objects/flocx_market_client.py
```diff
@@ -63,3 +64,65 @@ def send_offer(self, data):
...
+                url1 = CONF.flocx_market.endpoint_override + '/contract/' + contract_id
+                r_contract = self.get(url1)
+                c = json.loads(r_contract.content)
+                contract_data["contract_id"] = str(c["contract_id"])
+                # contract_data["time_created"] = str(c["time_created"])
+                contract_data["start_time"] = str(c["start_time"])
+                contract_data["end_time"] = str(c["end_time"])
+                contract_data["status"] = str(c["status"])
+                contract_data["cost"] = str(c["cost"])
+                contract_data["bid_id"] = str(c["bid_id"])
```
larsks: Why are we calling `str()` on all of these attributes? If this transformation is necessary, it should probably be in a loop rather than a series of individual statements.

esi_leap/tests/objects/test_contract.py
```diff
@@ -162,3 +212,39 @@ def test_save(self):
...
+        with mock.patch.object(flocx_market_client.FlocxMarketClient,
+                               'get_contracts', autospec=True) \
+            as mock_get_contracts:
+            mock_get_contracts.return_value = []
+            contracts = contract.Contract.retrieve_from_flocx_market()
+            mock_get_contracts.assert_called_once()
+            self.assertEqual(len(contracts), 0)
+            self.assertEqual([], contracts)
+
+    # def test_split(self):
```
larsks: Ditto here re: removing commented-out code.

DanNiESh: fixed~

esi_leap/objects/contract.py
DanNiESh: Yep!

esi_leap/objects/flocx_market_client.py
DanNiESh: Yeah, I moved get_contract functions to common/flocx_market.py. flocx_market_client.py does not have any contract related functions. I also created get_contract_by_id, get_ocr_by_status, get_offer_by_id in common/flocx_market.py. The code for retrieving contracts seems easier to understand :D

DanNiESh:
Another bug found:
https://github.com/DanNiESh/esi-leap/blob/173b072a638939758297c7f1fcca09ed16d69f8c/esi_leap/common/flocx_market.py#L37
I accidently put this line inside "if" block below, so new contracts are always created with the same offer id and ocr id. The updated PR should work.

esi_leap/objects/flocx_market_client.py
DanNiESh: I was trying to modify the string data got rom json.load() from unicode to utf-8. Now I add a object_hook to do the convert. https://github.com/DanNiESh/esi-leap/blob/cd637868eaecdb30fc59538a34a1a0bb4ea78a04/esi_leap/common/flocx_market.py#L12

esi_leap/manager/service.py
DanNiESh: Done!

esi_leap/objects/flocx_market_client.py
DanNiESh: Ah, yep! I've changed it to return [] when r.status_code != 200
https://github.com/DanNiESh/esi-leap/blob/60f70babd525ad00915ed879c2f59fc4bf6aee5c/esi_leap/common/flocx_market.py#L84

esi_leap/common/flocx_market.py
DanNiESh: Okay. I've set the new contract status to statuses.OPEN.

esi_leap/common/flocx_market.py
DanNiESh: I just pushed the code to have another query to bid.

esi_leap/objects/flocx_market_client.py
DanNiESh: Okay, I've changed the code a bit. Now provider queries marketplace very straightforward: get the offer_contract_pair, make get request to contract, offer, bid and create the contract. I was thinking to form the contract data in the structure that a single query should provide, like, using a single query provider can get the contracts with multiple offers, and it needs to split the contracts to create new ones. But for now provider skip the 'split contract' step.

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,110 @@
...
+from keystoneauth1 import adapter
+from keystoneauth1 import loading as ks_loading
+from oslo_config import cfg
+import datetime
+
+CONF = cfg.CONF
+
+def encode_hook(pairs):
+    new_pairs = {}
+    for key, value in pairs.iteritems():
```
tzumainn: python3 doesn't have `iteritems`, only `items`

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,110 @@
...
+        if isinstance(value, unicode):
```
tzumainn: this also doesn't work in python3

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,110 @@
...
+    res = marketplace_client.get(url)
+    return json.loads(res.content, object_hook=encode_hook)
+
+def update_contract(o_c_r_id, marketplace_client):
+    url = CONF.flocx_market.endpoint_override + \
+    '/offer_contract_relationship/' + o_c_r_id
+    data = {"status":"retrieved"}
+    r = marketplace_client.put(url, data)
+    if r.status_code != 200:
+        LOG.warning(
```
tzumainn: I don't think `LOG` is defined in this file?

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,110 @@
...
+
+def encode_hook(pairs):
+    new_pairs = {}
+    for key, value in pairs.iteritems():
+        if isinstance(value, unicode):
+            value = value.encode('utf-8')
+        if isinstance(key, unicode):
+            key = key.encode('utf-8')
+        new_pairs[key] = value
+    return new_pairs
```
tzumainn: Actually, I'm not sure you need this encode_hook at all; the code fails for me unless I take it out. Do you get an error without this?

esi_leap/common/flocx_market.py
```diff
@@ -0,0 +1,110 @@
...
+    if len(contract_offer_list) == 0:
+        return []
+    for contract_offer in contract_offer_list:
+        contract_data = {}
+        contract_id = contract_offer["contract_id"]
+        marketplace_offer_id = contract_offer["marketplace_offer_id"]
+        ocr_id = contract_offer["offer_contract_relationship_id"]
+        contract_data["marketplace_offer_contract_relationship_id"] = ocr_id 
+        c = get_contract_by_id(contract_id, marketplace_client)
+        contract_data["start_time"] = c["start_time"]
```
tzumainn: This should be `contract_data["start_date"] = c["start_time"]`; we have different column names, apparently. Same for the `end_time` on the next line.

esi_leap/db/sqlalchemy/api.py
tzumainn: We should probably just remove these commented out lines.

esi_leap/common/flocx_market.py
DanNiESh: Oh, I find that it's because my virtual environment use python2. Now I change it to python3 and take the encode_hook out.

esi_leap/common/flocx_market.py
DanNiESh: My bad! Fixed.

esi_leap/objects/offer.py
```diff
@@ -122,9 +121,9 @@ def to_marketplace_dict(self):
         offer_dict['end_time'] = offer_dict.pop('end_date').isoformat()
         offer_dict['cost'] = offer_dict['properties'].get('floor_price', 0)
         offer_dict['server_id'] = offer_dict.pop('resource_uuid')
-        offer_dict['provider_id'] = offer_dict.pop('uuid')
-        offer_dict['creator_id'] = offer_dict.pop('project_id')
-
+        offer_dict['provider_offer_id'] = offer_dict.pop('uuid')
+        offer_dict['creator_id'] = offer_dict['project_id']
```
jdtzmn: The `creator_id` is no longer necessary when creating offers in the marketplace

DanNiESh: Okay, I've removed it.

tzumainn:
Great job!

PR #13, opened 2020-06-05
Removed flocx dependencies

ljmcgann:
Removed all references and dependencies to the flocx marketplace
service. No longer requires flocx instance to create an offer.


ljmcgann:
There is one thing I am not sure about with this pr. When I cloned the repo and tried to run tox in a fresh venv, I got an error relating to the oslo.concurrency dependency. The culprit was a line #  -c{env:UPPER_CONSTRAINTS_FILE:https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt}
in the tox.ini. I commented out the line and was able to do the work just fine, but if having these requirements is a necessity, then something unrelated to removing flocx from esi-leap is wrong

DanNiESh:
> Removed all references and dependencies to the flocx marketplace
> service. No longer requires flocx instance to create an offer.

esi-leap has a flocx client to retrieve new contracts from the marketplace and creates it's own contracts in esi-leap db. I think esi-leap doesn't require a flocx server to create an offer.

tzumainn:
> > Removed all references and dependencies to the flocx marketplace
> > service. No longer requires flocx instance to create an offer.
> 
> esi-leap has a flocx client to retrieve new contracts from the marketplace and creates it's own contracts in esi-leap db. I think esi-leap doesn't require a flocx server to create an offer.

I think we should just remove all references to flocx. In retrospect, it was not great design to have flocx dependencies in esi-leap; if flocx needed esi information, it should have just pulled it. With development on esi-leap starting up again, I think it's a good idea for the repository to be as "clean" as possible.

tox.ini
```diff
@@ -13,7 +13,7 @@ setenv = VIRTUAL_ENV={envdir}
          PYTHONWARNINGS=default::DeprecationWarning
          TESTS_DIR=./esi_leap/tests/
 deps =
-  -c{env:UPPER_CONSTRAINTS_FILE:https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt}
+#  -c{env:UPPER_CONSTRAINTS_FILE:https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt}
```
tzumainn, line 16: Weirdly, I didn't get any tox errors with this line - what tox command are you running?

I checked a few openstack codebases, and the value they actually use is:

  -c{env:UPPER_CONSTRAINTS_FILE:https://releases.openstack.org/constraints/upper/master}

It may be worth changing ours to that as well?

ljmcgann: i was just using tox and tox -r in a fresh venv

DanNiESh: I don't have this issue either.

tzumainn:
Looks good - thanks!

PR #14, opened 2020-06-08
initial draft of esi-leap requirement doc

DanNiESh:
This requirement doc includes current logic of esi-leap: offer, contract and manager service. There should be more requirements to be proposed in future. Suggestion needed to know if it is on a right track and how to expand it further.

Thanks!
Danni


docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,15 @@
+# MOC esi-leap Requirements
+
+## Overview
+esi-leap is the leasing part of [Elastic Secure Infrastructure](https://github.com/CCI-MOC/esi) (esi), an OpenStack service for acting as a FLOCX provider to communicate with a FLOCX marketplace. It is intended to work on top of the [esi-common library](https://github.com/CCI-MOC/esi-common), which supports an OpenStack install that simulates Ironic multi-tenancy through the use of `project_owner_id` and `project_id` property attributes.
```
ljmcgann, line 4: esi-leap is its own complete Openstack service, not just a part of esi, as far as I understand it

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Definitions
+
+### Offer
+Baremetal node owners can offer up their nodes to others with nodes configurations, leasing period, and the price. On behalf of the node owners, esi-leap sends the offers to a FLOCX marketplace. An offer can be created, destroyed, saved, retrieved, and expired by esi-leap.
```
ljmcgann, line 9: It is not supposed to automatically send offers to FLOCX marketplace. I currently have a pending pr to fix this issue.

ljmcgann: But we could mention (perhaps in the overview?) that a marketplace service could be built on top of esi-leap. But regardless I feel like introducing the marketplace service here could be confusing

docs/esi-leap-requirements.md
ljmcgann: Also from the documentation I have read, esi-leap should be stylized ESI-Leap

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Contract
+A contract can allow nodes to be leased to a user within a specified period. esi-leap can create, destroy, save, retrieve, and expire a contract similar to an offer. Moreover, esi-leap is responsible for communicating with Ironic to fulfill the resources in the contracts.
+
+### Manager Service
```
ljmcgann, line 14: Same issues with flocx apply here.

docs/esi-leap-requirements.md
DanNiESh: Got it!

docs/esi-leap-requirements.md
DanNiESh: Yes, FLOCX could be confusing here. If your pr merges, I can make changes according to the newest repo.

docs/esi-leap-requirements.md
tzumainn: Agreed - let's talk about esi-leap as an independent leasing service first, and mention the connection to esi later.

esi-common *should* also no longer be needed; now Ironic nodes have 'owner' and 'lessee' attributes that replace the esi-common solution.

docs/esi-leap-requirements.md
tzumainn: The remove-flocx PR has merged, so feel free to remove any references here!

tzumainn: As this a requirements doc, it may also make sense to explicitly list offer fields (and the same with contract below).

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,15 @@
...
+An esi-leap manager has periodic jobs to manage offers and contracts. Firstly, it retrieves and checks for new contracts from a FLOCX market. If new contracts are retrieved, it then fulfills the resources in the contracts and updates the status of the contracts to 'FULFILLED'. The second job is periodically expiring offers and contracts that are out-of-date. The manager will update these with an 'EXPIRED' status.
```
tzumainn, line 15: I might suggest changing this paragraph into a list:

esi-leap manager has periodic jobs etc etc:

* expire offers
* fulfill contracts
* expire contracts

Lists are often an easier way for a reader to quickly parse meaning.

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,47 @@
...
+    {
+        'id': 27,
+        'uuid': '534653c9-880d-4c2d-6d6d-f4f2a09e384',
+        'project_id': '01d4e6a72f5c408813e02f664cc8c83e',
+        'resource_type': 'dummy_node',
+        'resource_uuid': '1718',
+        'start_date': datetime.datetime(2016, 7, 16, 19, 20, 30),
+        'end_date': datetime.datetime(2016, 8, 16, 19, 20, 30),
+        'status': statuses.AVAILABLE,
+        'properties': {'floor_price': 3},
```
tzumainn, line 20: Let's remove references to "price" - maybe just have this be an empty dict.

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,47 @@
+# ESI Leap Requirements
+
+## Overview
+ESI Leap is a leasing service for bare metal. Ironic nodes owners can offer up their nodes to other users, while leasees can view offers and create a contract for available nodes.
```
tzumainn, line 4: "lessees" instead of "leasees"

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,47 @@
...
+        'created_at': None,
+        'updated_at': None
+    }
+```
+
+### Manager Service
+An ESI Lesp manager has periodic jobs to manage offers and contracts. 
+* expire offers: out-of-date offers will be updated with an 'EXPIRED' status.
+* fulfill contracts: if new contracts are retrieved, the manager service will fulfill the resources in the contracts and updates the status of the contracts to 'FULFILLED'.
+* expire contracts: same as 'expire offers', ESI Leap will expire contracts based on timestamp.
```
tzumainn, line 47: In these examples, can you reference the actual fields - 'start_date' and 'end_date'?

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,47 @@
...
+## Definitions
+
+### Offer
+Ironic owners can bring up their nodes to ESI Leap with an offer. An offer consists of nodes configurations, leasing period, and the price, etc. An offer can be created, destroyed, saved, retrieved, and expired by ESI Leap. Here is an example offer schema:
+```
+    {
+        'id': 27,
+        'uuid': '534653c9-880d-4c2d-6d6d-f4f2a09e384',
+        'project_id': '01d4e6a72f5c408813e02f664cc8c83e',
+        'resource_type': 'dummy_node',
```
tzumainn, line 15: Let's use 'ironic_node' in the example.

docs/esi-leap-requirements.md
```diff
@@ -0,0 +1,47 @@
+# ESI Leap Requirements
+
+## Overview
+ESI Leap is a leasing service for bare metal. Ironic nodes owners can offer up their nodes to other users, while leasees can view offers and create a contract for available nodes.
+
+## Definitions
+
+### Offer
+Ironic owners can bring up their nodes to ESI Leap with an offer. An offer consists of nodes configurations, leasing period, and the price, etc. An offer can be created, destroyed, saved, retrieved, and expired by ESI Leap. Here is an example offer schema:
```
tzumainn, line 9: The offer doesn't really specify a node configuration; it specifies a node (through the resource_uuid). I think we should also eliminate references to 'price'.

DanNiESh:
Hi Mainn, I've made some changes according to your suggestions. Feel free to make any further comments!

PR #20, opened 2020-06-17
Add test_types.py

cudkowicz:
Adds initial tests the for the types file.

### DIFF
--- a/esi_leap/tests/api/controllers/test_types.py
+++ b/esi_leap/tests/api/controllers/test_types.py
@@ -1,0 +1,43 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from esi_leap.api.controllers import types as types
+import mock as mock
+import unittest as unittest
+from wsme import types as wtypes
+
+
+class TestCollection(unittest.TestCase):
+
+    def setUp(self):
+        self.test_collection = types.Collection()
+        self.test_collection._type = 'stuff'
+        self.test_collection.stuff = []
+        self.obj1 = mock.Mock(uuid='aaaaa')
+        self.obj2 = mock.Mock(uuid='bbbbb')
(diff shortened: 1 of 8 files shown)


esi_leap/tests/api/controllers/test_types.py
```diff
@@ -12,15 +12,35 @@
...
 
 
 class TestCollection(unittest.TestCase):
 
     def setUp(self):
         self.test_collection = types.Collection()
         self.test_collection._type = 'stuff'
-        self.test_collection.stuff = [1, 2, 3]
+        self.test_collection.stuff = []
+        self.test_collection.stuff.append(Tester('1', 'aaa'))
```
tzumainn, line 31: We probably don't need to create a Tester object class. Use mock instead:
```
   import mock


   self.test_collection.stuff.append(mock.Mock(uuid='aaa'))
```

esi_leap/tests/api/controllers/test_types.py
```diff
@@ -12,15 +12,35 @@
...
+        self.test_collection.stuff.append(Tester('2', 'aaaa'))
+        self.test_collection.stuff.append(Tester('3', 'aaaaa'))
```
tzumainn, line 33: Using variations of 'aaaa' isn't all that readable. I'd use different characters - aaaaa, bbbbb, ccccc, etc

esi_leap/tests/api/controllers/test_types.py
```diff
@@ -12,15 +12,35 @@
...
+        self._type = 'url'
 
     def test_has_next(self):
         self.assertEqual(self.test_collection.has_next(3), True)
         self.assertEqual(self.test_collection.has_next(0), False)
+
+    def test_get_next(self):
+        kwargs = {"key1": "Things", "key2": "Stuff"}
+        link = ("{'key1': 'Things', 'key2': 'Stuff'}"
+                "/v1/stuff?limit=3&marker=aaaaa")
```
tzumainn, line 43: Using 'aaaaa' is a little dangerous; someone might change the uuid above and not realize this test would fail. I'd suggest the following:

* In the setUp function, have "self.obj1 = mock.Mock(...)", for all three objects, and self.test_collection=[self.obj1, etc]
* Replace aaaaa here with self.obj3.uuid

esi_leap/tests/api/controllers/test_types.py
cudkowicz: i changed it to mock.

esi_leap/tests/api/controllers/test_types.py
cudkowicz: i switched a, b, and c, each of length 5.

esi_leap/tests/api/controllers/test_types.py
cudkowicz: i changed it to separate objects and then replaced aaaaa for the object.uuid.

PR #27, opened 2020-06-22
Fixed start and end times

ljmcgann:
Changed start_date and end_date to start_time and end_time
to better reflect functionality. Made it so start_times
must come before end_times. Added filtering so user's can get offers and contracts which are
available during a specified time range.


esi_leap/common/exception.py
```diff
@@ -66,3 +66,8 @@ class ResourceNoPermission(ESILeapException):
 
 class ResourceTypeUnknown(ESILeapException):
     msg_fmt = _("%(resource_type)s resource type unknown.")
+
+
+class InvalidTime(ESILeapException):
```
tzumainn, line 71: InvalidTime is a little vague - can you change this to InvalidTimeRange?

esi_leap/api/controllers/v1/contract.py
```diff
@@ -60,12 +61,17 @@ def get_one(self, contract_uuid):
...
                          datetime.datetime, datetime.datetime, wtypes.text,
                          wtypes.text)
-    def get_all(self, project_id=None, start_date=None, end_date=None,
+    def get_all(self, project_id=None, start_time=None, end_time=None,
                 status=None, offer_uuid=None):
         request = pecan.request.context
         cdict = request.to_policy_values()
         policy.authorize('esi_leap:contract:get', cdict, cdict)
 
+        if start_time and end_time is None or end_time and start_time is None:
```
tzumainn, line 70: This conditional is a little hard to read - can you put in parentheses here (and below)?

```
    if (start_time and end_time is None) or (end_time and start_time is None):
```

esi_leap/tests/api/controllers/v1/test_offer.py
```diff
@@ -39,3 +40,9 @@ def test_one(self):
         o = create_test_offer(self.context)
         data = self.get_json('/offers')
         self.assertEqual(o.uuid, data['offers'][0]["uuid"])
+
+    def test_get_invalid_time(self):
+
+        start = datetime.datetime(2016, 7, 16, 19, 20, 30)
+
+        self.get_json('/offers?start_time=' + str(start), expect_errors=True)
```
tzumainn, line 48: We should probably test this slightly more, as in https://opendev.org/openstack/ironic/src/branch/master/ironic/tests/unit/api/controllers/v1/test_node.py#L455-L465

tzumainn:
Looks good now - thanks!

PR #29, opened 2020-06-24
Tests for offer and contract

cudkowicz:
These files provide initial testing for the offer and contract files.


tzumainn:
Thanks for the PR! Unfortunately, I'm getting pep8 and py37 errors when I run tox against your PR. Can you double check that?

PR #30, opened 2020-06-25
Policy on contract get_all

ljmcgann:
Implemented policies on contract get all. This commit extends
the functionality of contract get_all to allow filtering by
project id and ownership in order to accomodate owner and lessee
use cases. Implements policy on top of these new filters.

### DIFF
--- a/esi_leap/api/controllers/v1/utils.py
+++ b/esi_leap/api/controllers/v1/utils.py
@@ -1,0 +1,24 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import oslo_policy
+
+from esi_leap.common import policy
+
+
+def verify_admin(cdict):
+    try:
+        policy.authorize('is_admin', cdict, cdict)
+        return True
+
+    except oslo_policy.policy.PolicyNotAuthorized:
+        return False
--- a/esi_leap/common/policy.py
+++ b/esi_leap/common/policy.py
@@ -20,12 +20,15 @@ CONF = esi_leap.conf.CONF
 _ENFORCER = None
 
 default_policies = [
-    policy.RuleDefault('is_owner',
-                       'role:owner or role:esi_leap_owner',
-                       description='Owner API access'),
     policy.RuleDefault('is_admin',
                        'role:admin or role:esi_leap_admin',
                        description='Full read/write API access'),
+    policy.RuleDefault('is_owner',
+                       'role:owner or role:esi_leap_owner',
+                       description='Owner API access'),
+    policy.RuleDefault('is_lessee',
+                       'role:lessee or role:esi_leap_lessee',
+                       description='Lessee API access'),
 ]
 
 contract_policies = [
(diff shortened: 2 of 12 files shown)


esi_leap/api/controllers/v1/utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+        policy.authorize('esi_leap:contract:get_all_owned_view', cdict, cdict)
+
+        if owner is None:
+            owner = r_project_id
+        else:
+            if not is_admin:
+                if r_project_id != owner:
+                    raise exception.ProjectNoPermission(project_id=project_id)
+
```
tzumainn, line 54: These lines (48-53) would also make a good util function, and could probably be made generic so it can be used for any project_id value, and not just an owner - something like

```
def get_project_value(request, default_project):
     <replace owner with default_project, and r_project_id with request.project_id>

owner = get_project_value(request, owner)
```

esi_leap/api/controllers/v1/utils.py
```diff
@@ -0,0 +1,84 @@
...
+        if owner:
+            raise exception.InvalidOwnerCommand(view=view, owner=owner)
+
+        if project_id is None:
+            project_id = r_project_id
+        else:
+            if not is_admin:
+                if project_id != r_project_id:
+                    raise exception.ProjectNoPermission(project_id=project_id)
```
tzumainn, line 69: Ah, yep, this is something that could then be replaced by the new util function.

esi_leap/api/controllers/v1/utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+        possible_filters['owner'] = owner
+        possible_filters['project_id'] = project_id
+    else:
+
+        policy.authorize('esi_leap:contract:get_all_no_view', cdict, cdict)
+
+        if owner:
+            raise exception.InvalidOwnerCommand(view=view, owner=owner)
```
tzumainn, line 62: Now that I think about it, this feels like it should be fine. In fact, I'm wondering if there should be a view=owned option at all. Maybe that can be taken out and just replaced by the --owner  filter. Then the CLI could have the following logic:

```
   if parsed_args.owned:
       owner = however you get the CLI user's project
   <pass in the owner filter>
```
This reduces the needed get_all policies to two, and may simplify all this code, which becomes:

```
  if view == 'all':
      <check admin>
  else:
       <if project filter undefined, set it>
       <if owner filter defined, verify user is admin or that owner matches user's project>

  <perform get_all call with appropriate filters>
```
In effect, we just do a whole bunch of initial logic to set various filters, and then just use those filters.

esi_leap/common/policy.py
```diff
@@ -36,10 +39,25 @@
...
+        'rule:is_admin or rule:is_lessee',
+        'Retrieve all contracts owned by project_id',
+        [{'path': '/contracts', 'method': 'GET'}]),
+    policy.DocumentedRuleDefault(
+        'esi_leap:contract:get_all_owned_view',
+        'rule:is_admin or rule:is_owner',
+        'Retrieve all contracts with offer owner',
+        [{'path': '/contracts', 'method': 'GET'}]),
+    policy.DocumentedRuleDefault(
+        'esi_leap:contract:get_all_all_view',
```
tzumainn, line 57: Let's rename these policies to match what Ironic uses - esi_leap:contract:list for non-admins, and esi_leap:contract:list_all (assuming we get rid of the get_all_owned_view rule).

esi_leap/api/controllers/v1/utils.py
```diff
@@ -0,0 +1,84 @@
...
+import oslo_policy
+
+from esi_leap.common import exception
+from esi_leap.common import policy
+
+
+def contract_get_all_authorize_filters(cdict, r_project_id,
+                                       start_time=None, end_time=None,
+                                       status=None, offer_uuid=None,
+                                       project_id=None, view=None, owner=None):
```
tzumainn, line 22: I think the bulk of this method should rename in contract.py; however there are definitely some bits that can be broken out into util functions (mentioned below).

esi_leap/db/sqlalchemy/api.py
```diff
@@ -234,19 +233,23 @@ def contract_get(context, contract_uuid):
     return result
 
 
-def contract_get_all(context, filters):
-    query = model_query(context, models.Contract, get_session())
+def contract_get_all(filters):
```
tzumainn, line 236: Don't forget to update db/api.py to match this method signature.

esi_leap/tests/api/controllers/v1/test_contract.py
```diff
@@ -11,10 +11,57 @@
...
+from oslo_policy import policy
+import testtools
+
+from esi_leap.api.controllers.v1.contract import ContractsController
+from esi_leap.common import exception
 from esi_leap.objects import contract
 from esi_leap.tests.api import base as test_api_base
 
 
+adminContext = ctx.RequestContext()
```
tzumainn, line 24: It seems more consistent (and python-y) to call this admin_context

PR #32, opened 2020-06-26
Policy for remaining offer and contract api functions

ljmcgann:
Implements policy checking for all remaining offer and contract
api endpoints. Refactors checking of privileges to exist at the api
level as opposed to the database level. This is done to make to ensure
all privilege checking is done at the top level api level.


esi_leap/api/controllers/v1/contract.py
```diff
@@ -54,9 +54,12 @@ class ContractsController(rest.RestController):
     def get_one(self, contract_uuid):
         request = pecan.request.context
         cdict = request.to_policy_values()
-        policy.authorize('esi_leap:contract:get', cdict, cdict)
+        policy.authorize('esi_leap:contract:list', cdict, cdict)
```
tzumainn, line 57: It's weird to rename the policy check for fetching one contract to "list" instead of "get". If you think we only need one of the two, it should probably be 'get'.

esi_leap/api/controllers/v1/contract.py
```diff
@@ -54,9 +54,12 @@ class ContractsController(rest.RestController):
...
 
-        c = contract.Contract.get(request, contract_uuid)
+        filters = ContractsController._contract_get_one_filters(
+            cdict, request.project_id)
```
tzumainn, line 60: Are we really just checking ownership? If so, I think we should call out the check more explicitly (and in the delete function), by which I mean: retrieve the contract, and then call a function that checks permissions.

We also don't really want to apply a filter to a 'get_one' call; filtering really only applies to lists.

I also don't think this check needs to be a separate function; after getting the contract, you just need:

```
if not verify_admin and c.project_id != request.project_id:
    throw WhateverException
```

esi_leap/api/controllers/v1/utils.py
```diff
@@ -22,3 +24,14 @@ def verify_admin(cdict):
 
     except oslo_policy.policy.PolicyNotAuthorized:
         return False
+
+
+def verify_resource_ownership(cdict, offer_dict):
```
tzumainn, line 29: I think you probably just want to pass in `resource_object` and `project_id` instead of `offer_dict`; it's hard to understand what needs to be in `offer_dict`.

esi_leap/objects/offer.py
```diff
@@ -40,28 +40,28 @@ class Offer(base.ESILEAPObject):
     }
 
     @classmethod
-    def get(cls, context, offer_uuid):
-        db_offer = cls.dbapi.offer_get(context, offer_uuid)
+    def get(cls, context, offer_uuid, filters):
```
tzumainn, line 43: I don't think it makes sense for a single get to have filters applied to it.

esi_leap/api/controllers/v1/contract.py
ljmcgann: So the two policies would be get and list_al? Or just get and list?

ljmcgann: The reason I did this is because I don't see why the listing policy would be different than the policy for getting just one contract

esi_leap/api/controllers/v1/contract.py
ljmcgann: I wanted to remove all the checks for being an admin and protecting data to outside of the dbapi

esi_leap/api/controllers/v1/contract.py
tzumainn: Yep, it's definitely fair to just use one! But if you just use one, name it 'get' instead of 'list'.

esi_leap/api/controllers/v1/contract.py
tzumainn: Definitely - I'm not saying you need to put that check into the database. But you can retrieve the contract, and *then* check permissions in the REST API.

esi_leap/api/controllers/v1/contract.py
ljmcgann: so the two policies will be get and list_all? or should it be get and list

ljmcgann: just slightly confused

esi_leap/api/controllers/v1/contract.py
ljmcgann: That wouldnt be possible for the contract get. Because the contract returned will not say anything about the owner of the related offer

ljmcgann: Unless we did a second call to the dbapi to then get the related offer. Doing it this way is just one call

esi_leap/api/controllers/v1/contract.py
tzumainn: It should either be 'get' or 'list', or you could just use 'get' and remove 'list' completely.

tzumainn: In other words, if you're just using one policy rule, replace `esi_leap:contract:list` everywhere with `esi_leap:contract:get`

esi_leap/api/controllers/v1/contract.py
tzumainn: Having two calls is fine. If you apply a filter, you're essentially still having the db enforce permissions. If we're just going to enforce permissions in one place, it should be here at the REST API.

esi_leap/api/controllers/v1/contract.py
ljmcgann: With one policy, then anybody can just see all of the contracts? I thought we didn't want that

ljmcgann: The way things are coded, its hardcoded that if you want to allow everybody to see any contract, then they can't because of checks like request.project != project_id

esi_leap/api/controllers/v1/contract.py
ljmcgann: well the db doesnt have to be fed a filter to work. The db doesn't enforce any permissions, the api just enforces how the db is called

esi_leap/api/controllers/v1/contract.py
ljmcgann: I will remedy this by calling the policy rather than just raising the error at that point

esi_leap/api/controllers/v1/contract.py
ljmcgann: If by enforce, you mean do the dirty work of surreptitiously producing no data then I suppose the db is enforcing

tzumainn: Yep, that's exactly what I mean - if for whatever reason we decide to loosen restrictions later, it'd be useful to have all the controlling logic be at the REST API level.

esi_leap/api/controllers/v1/contract.py
tzumainn: Wasn't your change already just having one policy, using 'list'? I'm just suggesting that you use 'get' instead of 'list'.

ljmcgann: We currently have two policies for contracts

ljmcgann: list_all for admins, and list for owners and lessees

tzumainn: Whoops - what I meant is, instead of using 'list' and 'list_all', use 'get' and 'get_all'.

tzumainn: And perhaps rename `get_all` to `get_admin`?

esi_leap/api/controllers/v1/offer.py
```diff
@@ -103,7 +104,17 @@ def post(self, new_offer):
...
-        o = offer.Offer(**new_offer.to_dict())
+        offer_dict = new_offer.to_dict()
+
+        if 'project_id' in offer_dict:
+            if offer_dict['project_id'] != request.project_id:
+                policy.authorize('esi_leap:contract:offer_admin', cdict, cdict)
+        else:
+            offer_dict['project_id'] = request.project_id
+
+        OffersController.verify_resource_permission(cdict, offer_dict)
```
tzumainn, line 115: Use `self.verify_resource_permission`

esi_leap/api/controllers/v1/offer.py
```diff
@@ -114,4 +125,15 @@ def delete(self, offer_uuid):
         policy.authorize('esi_leap:offer:delete', cdict, cdict)
 
         o = offer.Offer.get(request, offer_uuid)
-        o.destroy(pecan.request.context)
+        OffersController.verify_resource_permission(cdict, o.to_dict())
```
tzumainn, line 128: Same here

esi_leap/api/controllers/v1/offer.py
```diff
@@ -114,4 +125,15 @@ def delete(self, offer_uuid):
...
+        o.destroy()
+
+    @staticmethod
+    def verify_resource_permission(cdict, offer_dict):
+        resource_type = offer_dict.get('resource_type')
+        resource_uuid = offer_dict.get('resource_uuid')
+        resource = ro_factory.ResourceObjectFactory.get_resource_object(
+            resource_type, resource_uuid)
+        if not resource.is_resource_admin(offer_dict['project_id']):
+            policy.authorize('esi_leap:resource_admin', cdict, cdict)
```
tzumainn, line 139: Can we just use the offer admin policy?

esi_leap/api/controllers/v1/offer.py
```diff
@@ -114,4 +125,15 @@ def delete(self, offer_uuid):
         policy.authorize('esi_leap:offer:delete', cdict, cdict)
 
         o = offer.Offer.get(request, offer_uuid)
-        o.destroy(pecan.request.context)
+        OffersController.verify_resource_permission(cdict, o.to_dict())
+
+        o.destroy()
+
+    @staticmethod
+    def verify_resource_permission(cdict, offer_dict):
```
tzumainn, line 133: Use `_verify_resource_permission`

esi_leap/api/controllers/v1/offer.py
ljmcgann: Is there a reason this can't be a static method?

tzumainn: No, you're right - ignore this.

PR #33, opened 2020-07-02
Added availabilities to offers

ljmcgann:
Added availabilities to offers so that user's may see
a list of times in which an offer is available to make contracts
on. Added contraints to contract creation that a contract must 1:
have an offer_uuid that links to an existing offer and 2: start/
end_times for a contract must fall within an offer's availability range.
Added filtering to grab offers by their availabilities.

### DIFF
--- a/esi_leap/api/controllers/v1/offer.py
+++ b/esi_leap/api/controllers/v1/offer.py
@@ -35,12 +35,16 @@ class Offer(base.ESILEAPBase):
     end_time = wsme.wsattr(datetime.datetime)
     status = wsme.wsattr(wtypes.text)
     properties = {wtypes.text: types.jsontype}
+    availabilities = wsme.wsattr([[datetime.datetime]], readonly=True)
 
     def __init__(self, **kwargs):
 
         self.fields = offer.Offer.fields
         for field in self.fields:
             setattr(self, field, kwargs.get(field, wtypes.Unset))
+
+        setattr(self, 'availabilities', kwargs.get('availabilities',
+                                                   wtypes.Unset))
 
 
 class OfferCollection(types.Collection):
@@ -59,13 +63,17 @@ class OffersController(rest.RestController):
         policy.authorize('esi_leap:offer:get', cdict, cdict)
 
         o = offer.Offer.get(request, offer_uuid)
-        return Offer(**o.to_dict())
+        o = OffersController._add_offer_availabilities(o)
+
+        return Offer(**o)
 
     @wsme_pecan.wsexpose(OfferCollection, wtypes.text, wtypes.text,
                          wtypes.text, datetime.datetime, datetime.datetime,
+                         datetime.datetime, datetime.datetime,
                          wtypes.text)
     def get_all(self, project_id=None, resource_type=None,
                 resource_uuid=None, start_time=None, end_time=None,
+                available_start_time=None, available_end_time=None,
                 status=None):
 
         request = pecan.request.context
@@ -74,9 +82,27 @@ class OffersController(rest.RestController):
 
         if (start_time and end_time is None) or\
            (end_time and start_time is None):
-            raise exception.InvalidTimeCommand(resource='an offer',
-                                               start_time=str(start_time),
-                                               end_time=str(end_time))
+            raise exception.InvalidTimeAPICommand(resource='an offer',
+                                                  start_time=str(start_time),
+                                                  end_time=str(end_time))
+
+        if start_time and end_time and\
+           end_time <= start_time:
+            raise exception.InvalidTimeAPICommand(resource='an offer',
+                                                  start_time=str(start_time),
+                                                  end_time=str(end_time))
+
+        if (available_start_time and available_end_time is None) or\
+           (available_end_time and available_start_time is None):
+            raise exception.InvalidAvailabilityAPICommand(
+                a_start=str(start_time),
+                a_end=str(end_time))
+
+        if available_start_time and available_end_time and\
+                available_end_time <= available_start_time:
+            raise exception.InvalidAvailabilityAPICommand(
+                a_start=available_start_time,
+                a_end=available_end_time)
 
         possible_filters = {
             'project_id': project_id,
@@ -85,6 +111,8 @@ class OffersController(rest.RestController):
             'status': status,
             'start_time': start_time,
             'end_time': end_time,
+            'available_start_time': available_start_time,
+            'available_end_time': available_end_time,
         }
 
         filters = {}
@@ -94,8 +122,10 @@ class OffersController(rest.RestController):
 
         offer_collection = OfferCollection()
         offers = offer.Offer.get_all(request, filters)
+
         offer_collection.offers = [
-            Offer(**o.to_dict()) for o in offers]
+            Offer(**OffersController._add_offer_availabilities(o))
+            for o in offers]
         return offer_collection
 
     @wsme_pecan.wsexpose(Offer, body=Offer)
@@ -116,7 +146,10 @@ class OffersController(rest.RestController):
 
         o = offer.Offer(**offer_dict)
         o.create(request)
-        return Offer(**o.to_dict())
+
+        o = OffersController._add_offer_availabilities(o)
+
+        return Offer(**o)
 
     @wsme_pecan.wsexpose(Offer, wtypes.text)
     def delete(self, offer_uuid):
@@ -128,6 +161,13 @@ class OffersController(rest.RestController):
         OffersController._verify_resource_permission(cdict, o.to_dict())
 
         o.destroy()
+
+    @staticmethod
+    def _add_offer_availabilities(o):
+        availabilities = o.get_availabilities()
+        o = o.to_dict()
+        o['availabilities'] = availabilities
+        return o
 
     @staticmethod
     def _verify_resource_permission(cdict, offer_dict):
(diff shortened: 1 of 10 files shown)


esi_leap/db/sqlalchemy/api.py
```diff
@@ -133,23 +133,139 @@ def offer_get_all(filters):
...
+    for o in query:
+        try:
+            offer_verify_availability(o, start_time=a_start, end_time=a_end)
+        except exception.OfferNotAvailable:
+            query = query.filter(models.Offer.uuid != o.uuid)
+
     return query
 
 
+def offer_get_availabilities(offer_ref):
```
tzumainn, line 158: Is there a reason not to do this entire function in `objects/offer.py`?

```
contracts = self.contracts ()    // if there isn't a method for this, there should be 
for contract in contracts:
    // construct list of availabilities
```

esi_leap/db/sqlalchemy/api.py
```diff
@@ -133,23 +133,139 @@ def offer_get_all(filters):
...
+
+        a = [[a[j], a[j + 1]] for j in range(0, len(a) - 1, 2)]
+
+    else:
+        a = [[offer_ref.start_time, offer_ref.end_time]]
+
+    return a
+
+
+def offer_verify_availability(offer_ref, start_time=None, end_time=None):
```
tzumainn, line 190: This feels like it can be done through a single filtered query on contracts - something like:
```
select * from contracts c where c.offer_id = _offer_id_ and ((c.start_time <= _start_time_ and c.end_time >= _start_time_) or (c.end_time >= _end_time_ and c.start_time <= _end_time_))
```

esi_leap/db/sqlalchemy/api.py
```diff
@@ -133,23 +133,139 @@ def offer_get_all(filters):
...
     offer_ref = models.Offer()
     values['uuid'] = uuidutils.generate_uuid()
 
-    if 'start_time' in values and 'end_time' in values and \
-       values['start_time'] is not None and values['end_time'] is not None and \
-       values['start_time'] >= values['end_time']:
+    if 'start_time' not in values:
+        values['start_time'] = datetime.datetime.now()
+    if 'end_time' not in values:
+        values['end_time'] = datetime.datetime.max
```
tzumainn, line 260: If we're defaulting `project_id` at the REST API level, then we should be defaulting these values there as well.

esi_leap/db/sqlalchemy/api.py
```diff
@@ -133,23 +133,138 @@ def offer_get_all(filters):
...
+        raise exception.InvalidAvailabilityCommand(a_start=a_start,
+                                                   a_end=a_end)
+
     query = query.filter_by(**filters)
 
     if start and end:
         query = query.filter((start >= models.Offer.start_time) &
                              (end <= models.Offer.end_time))
 
+    for o in query:
```
tzumainn, line 149: This only needs to be done if a_start and a_end are defined, right? Otherwise it shouldn't be run at all?

esi_leap/db/sqlalchemy/api.py
ljmcgann: This cannot! The problem is the start_times and end_times in the db are not availability start and end_times, but rather UN-availability start and end times. I tried to see if I could do this in the db but it seemed impossible

esi_leap/objects/offer.py
```diff
@@ -79,3 +82,12 @@ def expire(self, context=None):
         # expire offer
         self.status = statuses.EXPIRED
         self.save(context)
+
+
+@versioned_objects_base.VersionedObjectRegistry.register
+class Availability(base.ESILEAPObject):
```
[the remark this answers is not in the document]
ljmcgann: I updated the pr to remove this

esi_leap/api/controllers/v1/offer.py
```diff
@@ -85,17 +95,28 @@ def get_all(self, project_id=None, resource_type=None,
             'status': status,
             'start_time': start_time,
             'end_time': end_time,
+            'available_start_time': available_start_time,
+            'available_end_time': available_end_time,
         }
 
+        print(available_start_time)
```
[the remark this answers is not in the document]
ljmcgann: removed

esi_leap/db/sqlalchemy/api.py
tzumainn: Right - but finding a conflicting contract - like through the query above - and saying "if one exists, then this isn't available" - feel like it should work? Or am I missing something?

Also, start_time/end_time should always be specified for this function, right?

esi_leap/api/controllers/v1/offer.py
```diff
@@ -85,17 +95,28 @@ def get_all(self, project_id=None, resource_type=None,
...
                 filters[k] = v
 
         offer_collection = OfferCollection()
         offers = offer.Offer.get_all(request, filters)
+
+        def a(o):
+            availabilities = o.get_availabilities()
+            o = o.to_dict()
+            o['availabilities'] = availabilities
+            return o
```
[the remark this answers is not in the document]
ljmcgann: A static method in OffersController?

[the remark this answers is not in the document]
tzumainn: I could actually see this being a class method in `objects/offer.py`. In fact, could it even be a parameter in `get` and `get_all` - if the flag is set, grab the availability at the object layer and add it in?

esi_leap/db/sqlalchemy/api.py
ljmcgann: This method does two things. It verifies if a start and end time are valid, and it finds the start and end time for contracts that did not pass in both parameters explicitly. Originally I had it so if you passed in a start time but no end time for contract creation, it grabbed the first availabilities end time. But I realized it should instead return the end time of the availability slot in which start_time lives.

ljmcgann: The checking validity could be done in the db, but the returning times part could not I think.

esi_leap/api/controllers/v1/offer.py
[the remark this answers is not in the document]
ljmcgann: The availabilities is never set in the object layer because it cannot be set as a field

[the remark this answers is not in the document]
ljmcgann: For the flag part, are you proposing that the availabilities not be added to the offer unless a flag is set?

esi_leap/db/sqlalchemy/api.py
ljmcgann: This function gets used by other db functions so I didn't want the db calling object level code

esi_leap/db/sqlalchemy/api.py
tzumainn: Okay - those two functions should be split. At first I thought that maybe this entire function should be moved out of the db layer; then I realized you were treating the start_time/end_time validity check as a database integrity issue, which is definitely fair. So that part should stay here, and probably be simplified into the kind of database query I expressed above.

The other half - finding start/end times if unspecified - should be separated out, and allowed to function in the object layer. Honestly, I'm starting to lean towards cutting this functionality from this patch for being overly complex. At most, I'd suggest the following:

* if start_time is unspecified, set it to current_time in the REST API layer
* if end_time is unspecified, look for the earliest start_time > current_time from a related contract (this could be grabbing the first contract from a filtered contract query); set this at the REST API layer
* let your database start/end time validation deal with the rest

esi_leap/db/sqlalchemy/api.py
ljmcgann: I did this because defaulting project id is a policy protection issue and what happens to it varies based on policy, but here the times are a data integrity issue. The whole thing just won't work if an end_time is set to null

esi_leap/db/sqlalchemy/api.py
ljmcgann: youre correct. I added a check.

esi_leap/api/controllers/v1/offer.py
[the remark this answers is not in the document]
tzumainn: How about adding a to_dict class method to objects/offer.py that accepts that flag then?

I am proposing that availabilities not be added to the offer unless the flag is set, yep.

esi_leap/db/sqlalchemy/api.py
tzumainn: I think it's only used to determine default start/end times, correct? If we simplify that, we can move this out of the db layer.

esi_leap/db/sqlalchemy/api.py
tzumainn: Isn't there a database constraint that would also prevent that? If not, then we should just add 'nullable=false' to the appropriate fields in the database model.

esi_leap/db/sqlalchemy/api.py
ljmcgann: I mean the work for no start or end time is there. Why not keep it? I can move the validity checks to other functions in the db however.

tzumainn: If it's not doing anything in the current code base, we don't want it in there as extra clutter. I'd keep it in a branch for now.

tzumainn:
Just tested this, and it works great! There are a few test failures; once that's fixed, I think this is good to merge.

PR #34, opened 2020-07-08
Tests for Offer and Contract

cudkowicz:
Adds testing for the post method in the offer file.

### DIFF
--- a/esi_leap/tests/api/controllers/v1/test_contract.py
+++ b/esi_leap/tests/api/controllers/v1/test_contract.py
@@ -11,6 +11,7 @@
 #    under the License.
 
 import datetime
+import mock
 from oslo_context import context as ctx
 from oslo_policy import policy
 import testtools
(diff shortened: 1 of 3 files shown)


esi_leap/tests/api/controllers/v1/test_offer.py
```diff
@@ -27,6 +27,10 @@ def create_test_offer(context):
     return o
 
 
+def create_test_offer_data():
+    return {1: 'test', 2: 'offer'}
```
tzumainn, line 31: Ah, we probably want to simulate actual test offer data more closely - something like:
```
{
   "resource_type": "test_node",
   "resource_uuid": "some-uuid",
   "start_time": "some_time_string"
}
```

cudkowicz: Ill add this in for offer. Someone made a ton of changes to contract so I figured I would make sure this works and then sift through the changes in the test to see where I should add it and such.

esi_leap/tests/api/controllers/v1/test_contract.py
```diff
@@ -62,6 +63,14 @@
 random_project = 'randomid'
 
 
+def create_test_contract_data():
+    return {
+   "uuid": "test_uuid",
+   "offer_uuid": "some-uuid",
```
tzumainn, line 69: we should be consistent in how we write our fake uuids "-" or "_"

cudkowicz: oops that was a typo

esi_leap/tests/api/controllers/v1/test_offer.py
```diff
@@ -27,6 +27,15 @@ def create_test_offer(context):
     return o
 
 
+def create_test_offer_data():
+
+    return {   
+        "resource_type": "test_node",
+        "resource_uuid": "some-uuid",
```
tzumainn, line 34: Here too :)

cudkowicz:
This provides tests for the post method in both offer and contract by adding post_json to base.py and using post_json.

tzumainn:
Okay, the code looks good now - however I think it might not quite work. I'm getting both pep8 and py37 errors - you may want to check those out!

PR #35, opened 2020-07-08
Add travis ci

naved001:
I used [black](https://pypi.org/project/black/) to format some lines to be under 79 characters. Personally, I am okay with increasing the line length to be more than 79 characters (maybe 88). Thoughts?


tzumainn:
I think 79 is standard across OpenStack, so I'm fine with 79. Thanks!

PR #36, opened 2020-07-08
Rework destroy to cancel

ljmcgann:
Changed the way the DELETE requets on offers and contracts
function. Instead of deleting the object from the database,
we instead change the object's status to a new 'cancelled'
status. Additionally, cancelling an offer will cancel all
associated contracts and free up the availability on the offer.

### DIFF
--- a/esi_leap/db/sqlalchemy/api.py
+++ b/esi_leap/db/sqlalchemy/api.py
@@ -10,17 +10,16 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-import datetime
 import sys
 
 from oslo_config import cfg
 from oslo_db.sqlalchemy import session as db_session
 from oslo_log import log as logging
-from oslo_utils import uuidutils
 
 import sqlalchemy as sa
 
 from esi_leap.common import exception
+from esi_leap.common import statuses
 from esi_leap.db.sqlalchemy import models
 
 
@@ -119,7 +118,7 @@ class InequalityCondition(object):
 # Offer
 def offer_get(offer_uuid):
     query = model_query(models.Offer, get_session())
-    offer_ref = query.filter_by(uuid=offer_uuid).first()
+    offer_ref = query.filter_by(uuid=offer_uuid).one_or_none()
     if not offer_ref:
         raise exception.OfferNotFound(offer_uuid=offer_uuid)
 
@@ -146,7 +145,7 @@ def offer_get_all(filters):
         for o in query:
             try:
                 offer_verify_availability(o, a_start, a_end)
-            except exception.OfferNotAvailable:
+            except exception.OfferNoTimeAvailabilities:
                 query = query.filter(models.Offer.uuid != o.uuid)
 
     return query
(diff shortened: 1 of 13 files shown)


esi_leap/db/sqlalchemy/api.py
```diff
@@ -344,10 +351,13 @@ def contract_update(contract_uuid, values):
     return contract_ref
 
 
-def contract_destroy(contract_uuid):
```
tzumainn, line 347: I don't think we should remove these functions entirely; I'd leave these in the object and db layers.

ljmcgann: So we will have both the cancel and destroy functions? And just let DELETE call only the cancel?

tzumainn: Yep, exactly.

esi_leap/db/sqlalchemy/api.py
```diff
@@ -155,33 +154,41 @@ def offer_get_all(filters):
...
-        .filter(models.Contract.offer_uuid == offer_ref.uuid)
-        .all()
-    )
+
+    return c_query.with_entities(
+        models.Contract.start_time, models.Contract.end_time).\
+        join(models.Offer).\
+        order_by(models.Contract.start_time).\
+        filter(models.Contract.offer_uuid == offer_ref.uuid,
+               models.Contract.status == statuses.OPEN).all()
```
tzumainn, line 163: Ah - good catch!

esi_leap/objects/offer.py
```diff
@@ -78,9 +82,31 @@ def get_availabilities(self):
...
         db_offer = self.dbapi.offer_create(updates)
         self._from_db_object(context, self, db_offer)
 
+    def cancel(self):
+        self.status = statuses.CANCELLED
+        self.save(None)
+
+        contracts = self.dbapi.contract_get_all({'owner': self.uuid})
+        for c in contracts:
+            c.cancel()
```
tzumainn, line 108: I think we may want to cancel contracts before setting the status to cancelled; this way, if there's an error in canceling a contract for whatever reason, the offer is marked as still open.

esi_leap/objects/contract.py
```diff
@@ -48,9 +52,41 @@ def get_all(cls, context, filters):
...
+        self.dbapi.offer_verify_availability(related_offer,
+                                             updates['start_time'],
+                                             updates['end_time'])
+
         db_contract = self.dbapi.contract_create(updates)
         self._from_db_object(context, self, db_contract)
 
+    def cancel(self):
+        self.status = statuses.CANCELLED
+        self.save(None)
```
tzumainn, line 88: I think you have to go through some of the contract's `expire` logic; if the contract is currently - active? - then you'll want to unassign the resource, etc etc.

esi_leap/objects/offer.py
ljmcgann: I would also have to find a way to rollback the cancelling of the contracts that did get cancelled for whatever reason, wouldnt I?

ljmcgann: Seems like a whole can of worms potentially

tzumainn: Yep, and it's probably something we want to handle in a future update; in the meantime, I'd argue that if the contracts can't all be canceled, then the offer's status should not be updated.

esi_leap/objects/contract.py
ljmcgann: This has been addressed

esi_leap/objects/offer.py
ljmcgann: I have made this change.

esi_leap/db/sqlalchemy/api.py
ljmcgann: I have re-added the destroy functions.

PR #41, opened 2020-07-12
Offer and contract api no longer can set project_id

ljmcgann:
Updated the offer and contract api to disallow setting
the project_id. This is to make esi-leap more consistent
with other openstack services and to prevent vague ownership
conditions. project_id is now a read only variable in both apis.

### DIFF
--- a/esi_leap/api/controllers/v1/contract.py
+++ b/esi_leap/api/controllers/v1/contract.py
@@ -29,7 +29,7 @@ from esi_leap.objects import offer
 class Contract(base.ESILEAPBase):
 
     uuid = wsme.wsattr(wtypes.text, readonly=True)
-    project_id = wsme.wsattr(wtypes.text)
+    project_id = wsme.wsattr(wtypes.text, readonly=True)
     start_time = wsme.wsattr(datetime.datetime)
     end_time = wsme.wsattr(datetime.datetime)
     status = wsme.wsattr(wtypes.text, readonly=True)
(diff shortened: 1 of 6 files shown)


esi_leap/api/controllers/v1/contract.py
```diff
@@ -149,7 +143,7 @@ def _contract_get_all_authorize_filters(cdict, r_project_id,
             policy.authorize('esi_leap:contract:get', cdict, cdict)
 
             if owner:
-                if r_project_id != owner:
+                if cdict['project_id'] != owner:
```
tzumainn, line 146: If I understand the code correctly - can't a user in project X specify project Y for their project_id and owner filter, and avoid the policy check below?

tzumainn: ah, I got confused - nevermind, although in the future this refactoring should be in a different PR

esi_leap/api/controllers/v1/contract.py
```diff
@@ -158,9 +152,9 @@ def _contract_get_all_authorize_filters(cdict, r_project_id,
             else:
 
                 if project_id is None:
-                    project_id = r_project_id
+                    project_id = cdict['project_id']
                 else:
-                    if project_id != r_project_id:
+                    if project_id != cdict['project_id']:
```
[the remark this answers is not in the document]
ljmcgann: I have addressed this

PR #42, opened 2020-07-13
Tests for api.py 

cudkowicz:
Added tests to api for both possibilities of the method offer_get, contract_get, and offer_update.

### DIFF
--- a/esi_leap/tests/db/sqlalchemy/test_api.py
+++ b/esi_leap/tests/db/sqlalchemy/test_api.py
@@ -100,7 +100,16 @@ test_contract_4 = dict(
     name='c2',
     start_time=now + datetime.timedelta(days=85),
     end_time=now + datetime.timedelta(days=90),
+    properties={},
     status=statuses.CANCELLED,
+)
+
+test_contract_5 = dict(
+    project_id='1e5533',
+    start_time=now + datetime.timedelta(days=90),
+    end_time=now + datetime.timedelta(days=100),
+    uuid='contract_5',
+    properties={},
 )
 
 
@@ -226,21 +235,16 @@ class TestAPI(base.DBTestCase):
         api.offer_verify_contract_availability(offer, start, end)
 
     def test_offer_get_by_uuid(self):
-
         offer = api.offer_create(test_offer_1)
-        api.offer_create(test_offer_2)
-
         res = api.offer_get_by_uuid(offer.uuid)
         self.assertEqual(offer.uuid, res.uuid)
         self.assertEqual(offer.project_id, res.project_id)
         self.assertEqual(offer.properties, res.properties)
 
     def test_offer_get_by_uuid_not_found(self):
-
         assert api.offer_get_by_uuid('some_uuid') is None
 
     def test_offer_get_by_name(self):
-
         o1 = api.offer_create(test_offer_1)
         o2 = api.offer_create(test_offer_2)
         o3 = api.offer_create(test_offer_3)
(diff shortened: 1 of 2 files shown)


esi_leap/tests/db/sqlalchemy/test_api.py
```diff
@@ -259,4 +266,122 @@ def test_offer_get_by_name(self):
 
     def test_offer_get_by_name_not_found(self):
 
-        assert api.offer_get_by_uuid('some_uuid') is None
+        self.assertEqual(api.offer_get_by_name('some_name'), [])
+
+    def test_offer_destroy(self):
+
```
tzumainn, line 272: You don't actually need this blank line after each function definition (I think you probably saw one early on and thought it was the standard).

esi_leap/tests/db/sqlalchemy/test_api.py
```diff
@@ -259,4 +266,122 @@ def test_offer_get_by_name(self):
...
+        test_contract_1['offer_uuid'] = o1.uuid
+        test_contract_2['offer_uuid'] = o2.uuid
+        test_contract_3['offer_uuid'] = o3.uuid
+        test_contract_4['offer_uuid'] = o4.uuid
+        c1 = api.contract_create(test_contract_1)
+        c2 = api.contract_create(test_contract_2)
+        c3 = api.contract_create(test_contract_3)
+        api.contract_create(test_contract_4)
+        res = api.contract_get_by_name('c1')
+        assert len(res) == 3
```
tzumainn, line 327: I was confused why you created so many contracts for this, but you're right, this function can return multiple results. Good thinking!

esi_leap/tests/db/sqlalchemy/test_api.py
```diff
@@ -259,4 +266,122 @@ def test_offer_get_by_name(self):
...
+        self.assertEqual(api.offer_get_by_uuid('offer_2'), None)
+
+    def test_offer_destroy_not_found(self):
+
+        self.assertEqual(api.offer_get_by_uuid('offer_4'), None)
+
+    def test_offer_update(self):
+
+        offer = api.offer_create(test_offer_3)
+        o = api.offer_create(test_offer_2)
```
tzumainn, line 284: You don't need to actually create this offer; you can just use test_offer_2['start_time'] below.

esi_leap/tests/db/sqlalchemy/test_api.py
```diff
@@ -259,4 +266,122 @@ def test_offer_get_by_name(self):
...
+                  'end_time': o.end_time}
+        api.offer_update(offer.uuid, values)
+        res = api.offer_get_by_uuid(offer.uuid)
+        self.assertEqual(res.start_time, offer.start_time)
+        self.assertEqual(res.end_time, offer.end_time)
+
+    def test_offer_get_all(self):
+
+        o = api.offer_create(test_offer_2)
+        offer = api.offer_create(test_offer_3)
```
tzumainn, line 295: I would use variable names like `o1` and `o2`; if the objects are similar, use similar variable names

esi_leap/tests/db/sqlalchemy/test_api.py
```diff
@@ -259,4 +266,122 @@ def test_offer_get_by_name(self):
...
+        res = api.offer_get_all({}).all()
```
tzumainn, line 296: Do you need the `all()` here?

esi_leap/tests/db/sqlalchemy/test_api.py
```diff
@@ -259,4 +266,122 @@ def test_offer_get_by_name(self):
...
+        self.assertEqual((o.to_dict(), offer.to_dict()),
+                         (res[0].to_dict(), res[1].to_dict()))
+
+    def test_contract_get_by_uuid(self):
+
+        o = api.offer_create(test_offer_2)
+        test_contract_4['offer_uuid'] = o.uuid
```
tzumainn, line 303: Ah, don't do this; in the test_contract_4 dict, hardcode the offer_uuid key to the offer uuid (likely test_offer_4's uuid).

esi_leap/tests/db/sqlalchemy/test_api.py
cudkowicz: I deleted all the blank lines after the functions defs.

esi_leap/tests/db/sqlalchemy/test_api.py
cudkowicz: ok ill change that around.

esi_leap/tests/db/sqlalchemy/test_api.py
cudkowicz: makes sense

esi_leap/tests/db/sqlalchemy/test_api.py
cudkowicz: i needed to create the offer in order to create the res to check the update against

esi_leap/tests/db/sqlalchemy/test_api.py
cudkowicz: nope!

esi_leap/tests/db/sqlalchemy/test_api.py
cudkowicz: ok ill fix that up!

esi_leap/tests/db/sqlalchemy/test_api.py
tzumainn: You need to create test_offer_3, but you don't need to create test_offer_2, right?

cudkowicz: oh i see now.

esi_leap/tests/db/sqlalchemy/test_api.py
cudkowicz: i remember having an issue with this working when we did it in the past. The problem was the uuid needed to belong to an offer that existed which was why we switched to doing it this way. Should I still cahnge it back?

tzumainn: Ah, you're right - leave it for now!

tzumainn:
Looks good - thanks!

PR #44, opened 2020-07-13
Changed owner url variable to set itself to project_id

ljmcgann:
The url variable when se to the value 'self' will set to the
project_id of the requester. This makes it so the user does
not have to enter their project_id every time they want to do
a related contracts request.

### DIFF
--- a/esi_leap/api/controllers/v1/contract.py
+++ b/esi_leap/api/controllers/v1/contract.py
@@ -149,11 +149,13 @@ class ContractsController(rest.RestController):
             policy.authorize('esi_leap:contract:get', cdict, cdict)
 
             if owner:
-                if r_project_id != owner:
-                    policy.authorize('esi_leap:contract:contract_admin',
-                                     cdict, cdict)
-
-                possible_filters['owner'] = owner
+                if owner == 'self':
+                    possible_filters['owner'] = cdict['project_id']
+                else:
+                    if r_project_id != owner:
+                        policy.authorize('esi_leap:contract:contract_admin',
+                                         cdict, cdict)
+                    possible_filters['owner'] = owner
                 possible_filters['project_id'] = project_id
             else:
 


ljmcgann:
The issue is leaving it unset vs setting it resuts in different behavior for not just owner, but also for project_id. You can only get contracts by your own project_id, unless you set the owner flag in which case project_id can be any value.

ljmcgann:
On the client side of things, I can simply do 'openstack lease contract list --owner' and get my own contracts, as I have set owner to 'self' when used without an argument. I initially was trying to do this independent of any changes to esi-leap, but on thinking about it it doesn't make much sense to have to send in your own project_id to esi-leap when it is easily grabbable from the context. Grabbing it from the client had me importing keystone clients to get a token, get a session, and getting a project_id. Just for this one default flag.

tzumainn:
> The issue is leaving it unset vs setting it resuts in different behavior for not just owner, but also for project_id. You can only get contracts by your own project_id, unless you set the owner flag in which case project_id can be any value.

Right - so this PR has unset -> all projects, project_id -> project_id ->, and 'self' -> self project, correct? Can't it be unset -> self project, project_id -> project_id, and 'all' -> all projects?

ljmcgann:
I think the issue with that would be if someone was both an owner and a lessee. If they want to get all of the contracts related to their offers, they would call owner=self with no project_id filter. They could set the project_id or not. If they want to get all the contracts they own, they intuitively would call by not setting project_id or owner, but then esi-leap will try to return all contracts you own that are attached to offers you made.

tzumainn:
> I think the issue with that would be if someone was both an owner and a lessee. If they want to get all of the contracts related to their offers, they would call owner=self with no project_id filter. They could set the project_id or not. If they want to get all the contracts they own, they intuitively would call by not setting project_id or owner, but then esi-leap will try to return all contracts you own that are attached to offers you made.

Ah, right - this is the owner filter, not the project_id filter. I think it's important to be consistent with how these filters behave; it helps for an intuitive understanding of the CLI. In general, I think we've defined empty -> self, with 'any/all' -> no filter. I kind of think we should do the same here.

ljmcgann:
Well currently we have not defined empty -> self. Empty owner implies nothing ever, empty project_id implies self for project_id only when no owner, and empty project_id implies nothing when owner is set.

tzumainn:
Ah, I see. I'm starting to think that we may be overloading the one CLI command to work for both owners and lessees. I wonder if it's more effective to have two different CLI commands for listing contracts - both of which still point to the same REST API 'get_all' - with slightly different filter options and defaults.

ljmcgann:
This wouldnt require any changes on esi-leap's end then right?

tzumainn:
Yep, exactly. It would be just two CLI commands allowing different filters.

tzumainn:
No longer needed!

PR #45, opened 2020-07-15
Update to manager

ljmcgann:
Fixed the manager calls to reflect changes to the object
layer code made in the last few commits.

### DIFF
--- a/esi_leap/manager/service.py
+++ b/esi_leap/manager/service.py
@@ -40,7 +40,6 @@ class ManagerService(service.Service):
         self._context = ctx.RequestContext(
             auth_token=None,
             project_id=None,
-            is_admin=True,
             overwrite=False)
 
     def start(self):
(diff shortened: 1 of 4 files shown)


ljmcgann:
Mainn I am trying to figure out the purpose of the _retrieve_contracts thread that gets created.

esi_leap/manager/service.py
```diff
@@ -63,28 +60,29 @@ def stop(self):
 
     def _fulfill_contracts(self):
         LOG.info("Checking for contracts to fulfill")
-        contracts = contract.Contract.get_all_by_status(
-            self._context, statuses.CREATED)
+        contracts = contract.Contract.get_all(
+            self._context, {'status': [statuses.ACTIVE, statuses.CREATED]})
```
tzumainn, line 64: If a contract is active, I don't think we need to fulfill it again.

esi_leap/manager/service.py
```diff
@@ -63,28 +60,29 @@ def stop(self):
...
+        now = timeutils.utcnow()
         for c in contracts:
-            if c.start_time and \
-               c.start_time <= timeutils.utcnow():
+            if c.start_time >= now and c.end_time <= now:
```
tzumainn, line 67: Isn't the opposite of what we want? This seems to be an impossible condition:
```
now -- start time ---------- end time -- now
```
We want:
```
start time -----now------end time
```
Which means start_time <= now and now <= end_time

ljmcgann: My bad. I did do a test run before but I mustve changed it before making the commit

esi_leap/manager/service.py
ljmcgann: fixed

esi_leap/manager/service.py
ljmcgann: fixed

tzumainn:
Looks good!

PR #46, opened 2020-07-16
Added names for api objects

ljmcgann:
Added the 'name' column to offers and contracts. Names
will make object management easier for users as they no
longer need to use cumbersome uuids to manage their objects.

### DIFF
--- a/esi_leap/api/controllers/v1/contract.py
+++ b/esi_leap/api/controllers/v1/contract.py
@@ -11,6 +11,7 @@
 #    under the License.
 
 import datetime
+from oslo_policy import policy as oslo_policy
 import pecan
 from pecan import rest
 import wsme
@@ -28,6 +29,7 @@ from esi_leap.objects import offer
 
 class Contract(base.ESILEAPBase):
 
+    name = wsme.wsattr(wtypes.text)
     uuid = wsme.wsattr(wtypes.text, readonly=True)
     project_id = wsme.wsattr(wtypes.text, readonly=True)
     start_time = wsme.wsattr(datetime.datetime)
@@ -52,19 +54,15 @@ class ContractCollection(types.Collection):
 class ContractsController(rest.RestController):
 
     @wsme_pecan.wsexpose(Contract, wtypes.text)
-    def get_one(self, contract_uuid):
+    def get_one(self, contract_id):
         request = pecan.request.context
         cdict = request.to_policy_values()
         policy.authorize('esi_leap:contract:get', cdict, cdict)
 
-        c = contract.Contract.get(request, contract_uuid)
+        permitted = ContractsController._contract_get_authorized_contract(
+            contract_id, cdict)
 
-        if c.project_id != request.project_id:
-            o = offer.Offer.get(request.project_id, c.offer_uuid)
-            if o.project_id != request.project_id:
-                policy.authorize('esi_leap:contract:get_admin', cdict, cdict)
-
-        return Contract(**c.to_dict())
+        return Contract(**permitted.to_dict())
 
     @wsme_pecan.wsexpose(ContractCollection, wtypes.text,
                          datetime.datetime, datetime.datetime, wtypes.text,
@@ -102,19 +100,48 @@ class ContractsController(rest.RestController):
         return Contract(**c.to_dict())
 
     @wsme_pecan.wsexpose(Contract, wtypes.text)
-    def delete(self, contract_uuid):
+    def delete(self, contract_id):
         request = pecan.request.context
         cdict = request.to_policy_values()
         policy.authorize('esi_leap:contract:delete', cdict, cdict)
 
-        c = contract.Contract.get(request, contract_uuid)
-        if c.project_id != request.project_id:
-            o = offer.Offer.get(request.project_id, c.offer_uuid)
-            if o.project_id != request.project_id:
+        permitted = ContractsController._contract_get_authorized_contract(
+            contract_id, cdict)
+
+        permitted.cancel()
+
+    @staticmethod
+    def _contract_get_authorized_contract(contract_id, cdict):
+
+        contract_objs = contract.Contract.get(contract_id)
+        permitted = []
+        for c in contract_objs:
+            try:
+                ContractsController._contract_authorize_management(c, cdict)
+                permitted.append(c)
+            except oslo_policy.PolicyNotAuthorized:
+                continue
+
+            if len(permitted) > 1:
+                raise exception.ContractDuplicateName(name=contract_id)
+
+        if len(permitted) == 0:
+            raise exception.ContractNotFound(contract_id=contract_id)
+
+        return permitted[0]
+
+    @staticmethod
+    def _contract_authorize_management(c, cdict):
+
+        if c.project_id != cdict['project_id']:
+            try:
                 policy.authorize('esi_leap:contract:contract_admin',
                                  cdict, cdict)
-
-        c.cancel()
+            except oslo_policy.PolicyNotAuthorized:
+                o = offer.Offer.get_by_uuid(c.offer_uuid)
+                if o.project_id != cdict['project_id']:
+                    policy.authorize('esi_leap:offer:offer_admin',
+                                     cdict, cdict)
 
     @staticmethod
     def _contract_get_all_authorize_filters(cdict,
--- a/esi_leap/api/controllers/v1/offer.py
+++ b/esi_leap/api/controllers/v1/offer.py
@@ -11,6 +11,7 @@
 #    under the License.
 
 import datetime
+from oslo_policy import policy as oslo_policy
 import pecan
 from pecan import rest
 import wsme
@@ -28,6 +29,7 @@ from esi_leap.resource_objects import resource_object_factory as ro_factory
 
 class Offer(base.ESILEAPBase):
 
+    name = wsme.wsattr(wtypes.text)
     uuid = wsme.wsattr(wtypes.text, readonly=True)
     project_id = wsme.wsattr(wtypes.text, readonly=True)
     resource_type = wsme.wsattr(wtypes.text, mandatory=True)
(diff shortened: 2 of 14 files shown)


esi_leap/api/controllers/v1/contract.py
```diff
@@ -52,19 +54,29 @@ def __init__(self, **kwargs):
...
+                permitted.append(c)
+            except oslo_policy.PolicyNotAuthorized:
+                continue
 
-        return Contract(**c.to_dict())
+            if len(permitted) > 1:
+                raise exception.ContractDuplicateName(name=contract_id)
+
+        if len(permitted) == 0:
+            raise exception.ContractNotFound(contract_id=contract_id)
```
tzumainn, line 77: This is definitely a good change - having ContractNotFound instead of an authorization error prevents a potential hacker from trying to figure out existing names simply by randomly trying to `get` a bunch of them.

esi_leap/api/controllers/v1/contract.py
```diff
@@ -52,19 +54,29 @@ def __init__(self, **kwargs):
...
     @wsme_pecan.wsexpose(Contract, wtypes.text)
-    def get_one(self, contract_uuid):
+    def get_one(self, contract_id):
         request = pecan.request.context
         cdict = request.to_policy_values()
         policy.authorize('esi_leap:contract:get', cdict, cdict)
 
-        c = contract.Contract.get(request, contract_uuid)
+        c_id = contract.Contract.get(contract_id)
+        print(c_id)
```
tzumainn, line 63: I'm guessing this can be removed?

esi_leap/api/controllers/v1/contract.py
```diff
@@ -102,19 +114,41 @@ def post(self, new_contract):
...
+                permitted.append(c)
+            except oslo_policy.PolicyNotAuthorized:
+                continue
+
+            if len(permitted) > 1:
+                raise exception.ContractDuplicateName(name=contract_id)
+
+        if len(permitted) == 0:
+            raise exception.ContractNotFound(contract_id=contract_id)
+
```
tzumainn, line 137: This code - 122 to 136 - could probably be in its own function. I might recommend putting this all in _contract_authorize_management and possibly renaming that function.

esi_leap/db/sqlalchemy/api.py
```diff
@@ -116,15 +116,18 @@ def clauses(self, field):
 
 
 # Offer
-def offer_get(offer_uuid):
+def offer_get_uuid(offer_uuid):
```
tzumainn, line 119: I have a vague sense that the convention would be something more like `offer_get_by_uuid`. Also, this needs to be added to `esi_leap/db/api.py`

esi_leap/db/sqlalchemy/api.py
```diff
@@ -116,15 +116,18 @@ def clauses(self, field):
...
     query = model_query(models.Offer, get_session())
     offer_ref = query.filter_by(uuid=offer_uuid).one_or_none()
-    if not offer_ref:
-        raise exception.OfferNotFound(offer_uuid=offer_uuid)
-
     return offer_ref
 
 
+def offer_get_name(name):
```
tzumainn, line 125: `offer_get_by_name` (and below for the contract functions)

esi_leap/objects/contract.py
```diff
@@ -41,9 +42,26 @@ class Contract(base.ESILEAPObject):
     }
 
     @classmethod
-    def get(cls, context, contract_uuid):
-        db_contract = cls.dbapi.contract_get(contract_uuid)
-        return cls._from_db_object(context, cls(), db_contract)
+    def get_uuid(cls, contract_uuid, context=None):
```
tzumainn, line 45: This should definitely be `get_by_uuid` (same with `get_` methods below)

ljmcgann:
Is 35 characters an acceptable upper limit on the name length. I did this to avoid the cheeky situtation where an offer has no name, and another owner comes along and makes an offer with name set to the uuid of the other offer. The way I return objects in get() looks for uuids before names and if it finds a uuid then it returns without looking for names. So nothing bad would happen (I did this to avoid the situation where the owner of the no name offer cannot delete their offer because of 'duplicate name' error) but the owner of the name=uuid offer would not be able to delete by his name despite the name not being a duplicate.

ljmcgann:
We also avoid any other unforseen issues with name=uuid. That was just one example I thought of.

esi_leap/api/controllers/v1/contract.py
ljmcgann: I have removed this. Good catch

tzumainn:
> Is 35 characters an acceptable upper limit on the name length. I did this to avoid the cheeky situtation where an offer has no name, and another owner comes along and makes an offer with name set to the uuid of the other offer. The way I return objects in get() looks for uuids before names and if it finds a uuid then it returns without looking for names. So nothing bad would happen (I did this to avoid the situation where the owner of the no name offer cannot delete their offer because of 'duplicate name' error) but the owner of the name=uuid offer would not be able to delete by his name despite the name not being a duplicate.

Oh wow, I didn't even think about that. Yeah, I think that's a good idea!

esi_leap/api/controllers/v1/contract.py
```diff
@@ -102,19 +100,48 @@ def post(self, new_contract):
...
+        permitted = []
+        for c in c_id:
+            try:
+                ContractsController._contract_authorize_management(c, cdict)
+                permitted.append(c)
+            except oslo_policy.PolicyNotAuthorized:
+                continue
+
+            if len(permitted) > 1:
+                raise exception.ContractDuplicateName(name=contract_id)
```
tzumainn, line 126: I think it's fine that we only ever return one permitted contract here; however if so, I think the method should be renamed to remove `_all`, and we should just return `permitted[0]` right away.

esi_leap/api/controllers/v1/contract.py
```diff
@@ -102,19 +100,48 @@ def post(self, new_contract):
...
+        if c.project_id != cdict['project_id']:
+            try:
                 policy.authorize('esi_leap:contract:contract_admin',
                                  cdict, cdict)
-
-        c.cancel()
+            except oslo_policy.PolicyNotAuthorized:
+                o = offer.Offer.get_by_uuid(c.offer_uuid)
+                if o.project_id != cdict['project_id']:
+                    policy.authorize('esi_leap:contract:contract_admin',
```
tzumainn, line 143: Should this check the `offer_admin` policy?

esi_leap/api/controllers/v1/offer.py
```diff
@@ -153,15 +161,31 @@ def post(self, new_offer):
...
 
-        o = offer.Offer.get(request, offer_uuid)
-        OffersController._verify_resource_permission(cdict, o.to_dict())
+        o_id = offer.Offer.get(offer_id)
+
+        permission = []
+        for o in o_id:
+            try:
+                if o.project_id != request.project_id:
+                    policy.authorize('esi_leap:contract:offer_admin',
```
tzumainn, line 175: This should be 'offer:offer_admin'

esi_leap/api/controllers/v1/offer.py
```diff
@@ -58,13 +60,19 @@ def __init__(self, **kwargs):
...
         policy.authorize('esi_leap:offer:get', cdict, cdict)
 
-        o = offer.Offer.get(request, offer_uuid)
-        o = OffersController._add_offer_availabilities(o)
+        o_id = offer.Offer.get(offer_id)
+
+        if len(o_id) > 1:
+            raise exception.OfferDuplicateName(name=offer_id)
+        elif len(o_id) == 0:
+            raise exception.OfferNotFound(offer_uuid=offer_id)
```
tzumainn, line 73: I noticed that you check the `offer_admin` policy in `delete`, but not here - any reason why not? If you end up checking it here as well, then that logic can probably be encapsulated in a function, similarly to the contract code.

esi_leap/api/controllers/v1/contract.py
```diff
@@ -102,19 +100,48 @@ def post(self, new_contract):
...
-            if o.project_id != request.project_id:
+        permitted = ContractsController._contract_get_all_authorized_contracts(
+            contract_id, cdict)
+
+        permitted[0].cancel()
+
+    @staticmethod
+    def _contract_get_all_authorized_contracts(contract_id, cdict):
+
+        c_id = contract.Contract.get(contract_id)
```
tzumainn, line 116: Maybe call it `contracts` instead of `c_id`?

esi_leap/api/controllers/v1/contract.py
ljmcgann: Your'e correct. I have changed this.

esi_leap/api/controllers/v1/offer.py
ljmcgann: fixed

esi_leap/api/controllers/v1/offer.py
ljmcgann: The reason I do not check any policy here is because we had decided that anybody can view any offer.

ljmcgann: Obviously anybody should not be able to delete any offer randomly so I use policy in delete()

esi_leap/api/controllers/v1/contract.py
ljmcgann: I just didn't like the fact that there is an imported module named contract, we are in a module named contract, the imported module uses a class called Contract and the module we are in defines a class called Contract.

ljmcgann: @mock.patch('esi_leap.api.controllers.v1.offer.offer.Offer.create')

esi_leap/api/controllers/v1/offer.py
tzumainn: Ah, you're 100% correct!

esi_leap/api/controllers/v1/contract.py
tzumainn: > I just didn't like the fact that there is an imported module named contract, we are in a module named contract, the imported module uses a class called Contract and the module we are in defines a class called Contract.

Ah, fair enough. However, if `contract.Contract.get` is returning multiple contract objects, the assigned variable should almost definitely be plural, and drop the `_id`. Maybe `contract_objs`?

ljmcgann: Fixed

esi_leap/api/controllers/v1/contract.py
ljmcgann: fixed

tzumainn:
Looks good - thanks!

PR #47, opened 2020-07-24
Conflict times for offer creation

ljmcgann:
Added a check to offer creation to see if an offer
has a time conflict with another offer on the same
resource node. This is needed as having two offers
exist on the same node allows one node to be leased
out to multiple lessees concurrently.

### DIFF
--- a/esi_leap/common/exception.py
+++ b/esi_leap/common/exception.py
@@ -72,6 +72,13 @@ class OfferNoTimeAvailabilities(ESILeapException):
                 "time range %(start_time)s, %(end_time)s.")
 
 
+class OfferResourceTimeConflict(ESILeapException):
+    msg_fmt = ("Offer on %(resource_type)s %(resource_uuid)s cannot be "
+               "created with a start_time and end_time "
+               "which conflicts with the time range of an existing offer "
+               "on %(resource_type)s %(resource_uuid)s.")
+
+
 class OfferNotAvailable(ESILeapException):
     msg_fmt = _("Offer %(offer_uuid)s does not have status "
                 "'available'. Got offer status '%(status)s'.")
--- a/esi_leap/db/api.py
+++ b/esi_leap/db/api.py
@@ -109,9 +109,14 @@ def offer_get_conflict_times(offer_ref):
     return IMPL.offer_get_conflict_times(offer_ref)
 
 
-def offer_verify_availability(offer_ref, start, end):
-    return IMPL.offer_verify_availability(
+def offer_verify_contract_availability(offer_ref, start, end):
+    return IMPL.offer_verify_contract_availability(
         offer_ref, start, end)
+
+
+def offer_verify_resource_availability(r_type, r_uuid, start, end):
+    return IMPL.offer_verify_resource_availability(
+        r_type, r_uuid, start, end)
 
 
 def offer_create(values):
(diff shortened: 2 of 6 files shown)


esi_leap/common/exception.py
```diff
@@ -72,6 +72,13 @@ class OfferNoTimeAvailabilities(ESILeapException):
                 "time range %(start_time)s, %(end_time)s.")
 
 
+class OfferTimeConflictResourceNode(ESILeapException):
```
tzumainn, line 75: This should probably be renamed to just 'OfferResourceTimeConflict', since the data model never actually talks about node. I could see putting `%(resource_type)` in the error message: `("Offer on %(resource_type)s %(resource_uuid)s cannot be...")` This lets us potentially support other resource types in the future.

esi_leap/db/api.py
```diff
@@ -114,6 +114,11 @@ def offer_verify_availability(offer_ref, start, end):
         offer_ref, start, end)
 
 
+def offer_verify_node_availability(resource_node_uuid, start, end):
```
tzumainn, line 117: I'd use `resource_availability` and `resource_uuid`

esi_leap/db/api.py
```diff
@@ -114,6 +114,11 @@ def offer_verify_availability(offer_ref, start, end):
...
+    return IMPL.offer_verify_availability(
```
tzumainn, line 118: I think you meant the new verify function here.

esi_leap/db/sqlalchemy/api.py
```diff
@@ -216,6 +215,31 @@ def offer_verify_availability(offer_ref, start, end):
                                                   end_time=end)
 
 
+def offer_verify_node_availability(resource_node_uuid, start, end):
```
tzumainn, line 218: `offer_verify_resource_availability`; and maybe rename `offer_verify_availability` to `offer_verify_contract_availability`?

esi_leap/common/exception.py
ljmcgann: I changed the text. I cannot add the resource_type unless I want to pass it in as an extra parameter to the db function solely for the purpose of raising errors. It seems silly to me but if you think it is worth it then I can do so.

esi_leap/db/api.py
ljmcgann: resource_node_uuid makes more sense because we are passing in a uuid and not an availability.

esi_leap/db/api.py
ljmcgann: fixed

esi_leap/db/sqlalchemy/api.py
ljmcgann: fixed

esi_leap/common/exception.py
tzumainn: I think there's a misunderstanding here; a resource_uuid does not unique define a resource; it's a resource_type + resource_uuid. Nothing in the offer/contract code should directly reference a node, only a 'resource' which the resource factory class later instantiates. That probably means that `offer_verify_node_availability` needs to take in `resource_type` and filter based on resource_uuid and resource_type

ljmcgann: ah I did misunderstand it. Its a quick fix.

PR #48, opened 2020-07-24
Contract creation by name

ljmcgann:
Added the ability to use an offer's name in place
of uuid when creating a contract.

### DIFF
--- a/esi_leap/api/controllers/v1/contract.py
+++ b/esi_leap/api/controllers/v1/contract.py
@@ -12,6 +12,7 @@
 
 import datetime
 from oslo_policy import policy as oslo_policy
+from oslo_utils import uuidutils
 import pecan
 from pecan import rest
 import wsme
@@ -36,7 +37,8 @@ class Contract(base.ESILEAPBase):
     end_time = wsme.wsattr(datetime.datetime)
     status = wsme.wsattr(wtypes.text, readonly=True)
     properties = {wtypes.text: types.jsontype}
-    offer_uuid = wsme.wsattr(wtypes.text, mandatory=True)
+    offer_uuid = wsme.wsattr(wtypes.text, readonly=True)
+    offer_uuid_or_name = wsme.wsattr(wtypes.text)
 
     def __init__(self, **kwargs):
         self.fields = contract.Contract.fields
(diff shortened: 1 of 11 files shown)


esi_leap/api/controllers/v1/contract.py
```diff
@@ -95,7 +96,7 @@ def post(self, new_contract):
         contract_dict['project_id'] = request.project_id
 
         c = contract.Contract(**contract_dict)
-        c.create(request)
+        c.create(new_contract.offer_uuid_or_name, request)
```
tzumainn, line 99: Won't offer_uuid_or_name be part of the contract_dict? I don't think it needs to be separated out like this.

ljmcgann: It will not be.         contract_dict = new_contract.to_dict() willl turn the passed in data into a contract dictionary based on the fields it finds in objects/contract Contract object. So the offer_uuid_or_name will not be added.

ljmcgann: I could add it to the dictionary, but making the dictionary inconsistent seemed to violate the purpose of having the to_dict() function

ljmcgann: The to_dict() should be the dictionary representation of a Contract object. Adding more fields in seemed like a cheap way to circumvent it.

tzumainn: Ah, gotcha. In that case, it might make sense to process the offer_uuid_or_name here in the REST API, and not in the object layer.

ljmcgann: And would be confusing to anybody who is trying to call the object layer not through the pecan api (which is a thing we think about?)

ljmcgann: I originally had that, but it seemed inefficient. To do that I had to make a query to the db to get the related offer for the contract. But in the object layer create function, we also do that exact same call. It seemed pointless to go to the db twice for the same thing

tzumainn: It seems far messier to me to start adding extra parameters to a pretty clean create method. The object layer typically won't be directly exposed; instead, when Ironic needs to access, say, Neutron, it instantiates a neutron client - python-neutronclient - and uses that.

ljmcgann: So the cost of going to the db twice for the same exact object is worth it?

tzumainn: Wouldn't you just move the code you added to the object layer into the rest api layer?

ljmcgann: An alternative is moving the code that requires us to go to the db in the object layer to the api layer. But I don't really like this option as I think having the create function do all the work of turning data into something the db can process and setting any default values.

ljmcgann: The object create function already had to go to the db to get the relate offer to 1: verify it exists, 2: use it to set the default start and end time, and 3: check conflicts. This code could be moved to the api layer but it seemed more intuitive to have it in the object layer.

ljmcgann: I am now adding code that resolves an offer_uuid_or_name into an offer_uuid through a query to the db. If I put that in the create, I only have to go to the db once. If I put it in the api, unless I move all the other code out of the object layer, that will be two queries to the db for the same exact offer.

tzumainn: I think I would fundamentally prefer having a duplicated query than to have the object layer `create` method start having additional parameters. Adding a separate parameter such as `offer_uuid_or_name` changes the object layer from being a relatively thin layer over the database for use by things like the REST API and the manager service, into something a little more prescriptive solely for the sake of the REST API.

I took another look at the Ironic REST API/object layers for nodes. The api layer has a mix of validation and default value setting:

https://opendev.org/openstack/ironic/src/branch/master/ironic/api/controllers/v1/node.py#L2232

While the object layer limits itself to validations:

https://opendev.org/openstack/ironic/src/branch/master/ironic/objects/node.py#L372

It feels like you may be able to reduce a database query by changing `offer_get_first_availability` and `offer_verify_availability` to accept an offer_uuid instead of the entire offer object. The first doesn't need anything beyond the offer uuid; the second feels like you might be able to avoid needing offer_ref.start_time if you update the query into a join, and add a clause in the conflicts filter that checks the offer start/end times.

ljmcgann: I thought of that for those two function calls. But there was still the issue of the default value for the end time and the verifying that the offer exits. Obvioulsy if we have code that resolves an offer_uuid_or_name into an offer, we don't need to check that it exists again. And if you think the default values should be moved to the api, then we can eliminate the db query entirely.

tzumainn: Yep, I think moving the default values to the API may make sense here!

ljmcgann: I have done this. It uses two calls to the db however.

PR #50, opened 2020-07-29
Additional tests for Api.py 

cudkowicz:
Provides tests for the exceptions in many methods within api.py as well as a bug fix.

### DIFF
--- a/esi_leap/tests/db/sqlalchemy/test_api.py
+++ b/esi_leap/tests/db/sqlalchemy/test_api.py
@@ -234,12 +234,31 @@ class TestAPI(base.DBTestCase):
         end = now + datetime.timedelta(days=87)
         api.offer_verify_contract_availability(offer, start, end)
 
+    def test_offer_get_conflict_times(self):
+        o1 = api.offer_create(test_offer_1)
+        self.assertEqual(api.offer_get_conflict_times(o1), [])
+        test_contract_3['offer_uuid'] = o1.uuid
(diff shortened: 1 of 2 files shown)


cudkowicz:
I believe I added tests for all the methods in offer and contract now.

esi_leap/tests/db/sqlalchemy/test_api.py
```diff
@@ -234,12 +234,21 @@ def test_offer_verify_contract_availability(self):
...
 
+    def test_offer_get_conflict_times(self):
+        o1 = api.offer_create(test_offer_1)
+        self.assertEqual(api.offer_get_conflict_times(o1), [])
+
+    def test_offer_get_first_availability(self):
+        o1 = api.offer_create(test_offer_1)
+        self.assertEqual(api.offer_get_first_availability
+                         (o1.uuid, o1.start_time,), None)
+
```
tzumainn, line 245: I think I'd update these tests so that the tested functions return something beyond empty results. Just as a quick refresher: an offer is created with a start and end time; by default its availability is that start time and end time. Now let's say someone creates a contract from c.start to c.end:

START----------------------c.start----------c.end----------------------END

The new availabilities should be START->c.start and c.end->END.

I assume a conflict time would be the inverse - so c.start-->c.end.

tzumainn:
It looks like there's a failing test. Also, you should edit your commit message to remove the "WIP"

PR #51, opened 2020-07-29
Added locks to contract

ljmcgann:
Added locks using oslo concurrency to contracts.
Without this contracts were able to be expired multiple times.

### DIFF
--- a/esi_leap/manager/service.py
+++ b/esi_leap/manager/service.py
@@ -16,6 +16,8 @@ import esi_leap.conf
 from esi_leap.manager import utils
 from esi_leap.objects import contract
 from esi_leap.objects import offer
+
+from oslo_concurrency import lockutils
 from oslo_context import context as ctx
 from oslo_log import log as logging
 import oslo_messaging as messaging
@@ -75,8 +77,14 @@ class ManagerService(service.Service):
         now = timeutils.utcnow()
         for c in contracts:
             if c.end_time <= now:
-                LOG.info("Expiring contract %s", c.uuid)
-                c.expire(self._context)
+
+                with lockutils.lock(c.uuid,
+                                    lock_file_prefix='contract',
+                                    external=True):
+                    c_locked = contract.Contract.get_by_uuid(c.uuid)
+                    if c_locked.status != statuses.EXPIRED:
+                        c_locked.expire(self._context)
+                        LOG.info("Expiring contract %s", c.uuid)
 
     def _expire_offers(self):
         LOG.info("Checking for expiring offers")
--- a/esi_leap/objects/contract.py
+++ b/esi_leap/objects/contract.py
@@ -16,6 +16,7 @@ from esi_leap.db import api as dbapi
 from esi_leap.objects import base
 from esi_leap.objects import fields
 from esi_leap.objects.offer import Offer
+
 from oslo_config import cfg
 from oslo_versionedobjects import base as versioned_objects_base
 
@@ -112,8 +113,15 @@ class Contract(base.ESILEAPObject):
         self.save(context)
 
     def expire(self, context=None):
+
+        if self.status == statuses.EXPIRED:
+            raise Exception("Already expired " + self.uuid)
+
         # unassign resource
         o = Offer.get_by_uuid(self.offer_uuid, context)
+
+        if o.status != statuses.AVAILABLE:
+            raise Exception("Offer not available " + o.uuid)
 
         resource = o.resource_object()
         if resource.get_contract_uuid() == self.uuid:
--- a/esi_leap/objects/offer.py
+++ b/esi_leap/objects/offer.py
@@ -14,10 +14,11 @@
 from esi_leap.common import statuses
 from esi_leap.db import api as dbapi
 from esi_leap.objects import base
-import esi_leap.objects.contract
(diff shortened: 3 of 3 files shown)


esi_leap/objects/contract.py
```diff
@@ -112,9 +113,16 @@ def fulfill(self, context=None):
         self.save(context)
 
     def expire(self, context=None):
+
+        if self.status == statuses.EXPIRED:
+            raise Exception("Already expired " + self.uuid)
```
tzumainn, line 118: I don't think this is needed; if it's already expired, then just return.

esi_leap/objects/contract.py
```diff
@@ -112,9 +113,16 @@ def fulfill(self, context=None):
...
+
         # unassign resource
         o = Offer.get_by_uuid(self.offer_uuid, context)
 
+        if o.status != statuses.AVAILABLE:
+            raise Exception("Offer not available " + o.uuid)
+
```
tzumainn, line 125: I know you've brought up the reason for this before, but could you explain the need for this check again?

esi_leap/manager/service.py
```diff
@@ -75,8 +77,14 @@ def _expire_contracts(self):
...
-                LOG.info("Expiring contract %s", c.uuid)
-                c.expire(self._context)
+
+                with lockutils.lock(c.uuid,
+                                    lock_file_prefix='contract',
+                                    external=True):
+                    c_locked = contract.Contract.get_by_uuid(c.uuid)
+                    if c_locked.status != statuses.EXPIRED:
+                        c_locked.expire(self._context)
+                        LOG.info("Expiring contract %s", c.uuid)
```
tzumainn, line 87: Given that there are similar locks here, and in the offer expire method - would it make sense to have the lock be in the actual contract expire method?

ljmcgann:
I don't think you can use the decorator because we are using specific instances of the method from a class. If you go to the db, get a contract that is being expired, then go to the expire function and get blocked, that means that another process is currently performing the expire function. But the reference you grabbed out of the db will not have the changes reflecting that.

esi_leap/manager/service.py
ljmcgann: You cannot do this for the reason I stated below.

ljmcgann:
One of the downsides to doing things this way is the db queries. We need a first query to get all the uuid's of contract's which need to be expired. We use these uuids to get locks, then once we get the lock we have to go get the contract by its uuid. this means there are #contracts+1 visits to the db. The +1 is the getting all by offer_uuid+status (an expired contract can never become not expired so we can use this extra filter if we want to reduce the amount of contracts we might have to request locks for) and the remaining queries are a run to the db to get a guaranteed up to date contract. The first db run is going to be an O(n) query. But I think the other queries should be O(1) as we are trying to get by the primary key. Please correct me if I am wrong in this understanding. Otherwise, this system might be kinda slow.

tzumainn:
> One of the downsides to doing things this way is the db queries. We need a first query to get all the uuid's of contract's which need to be expired. We use these uuids to get locks, then once we get the lock we have to go get the contract by its uuid. this means there are #contracts+1 visits to the db. The +1 is the getting all by offer_uuid+status (an expired contract can never become not expired so we can use this extra filter if we want to reduce the amount of contracts we might have to request locks for) and the remaining queries are a run to the db to get a guaranteed up to date contract. The first db run is going to be an O(n) query. But I think the other queries should be O(1) as we are trying to get by the primary key. Please correct me if I am wrong in this understanding. Otherwise, this system might be kinda slow.

Well, it looks like we already filter the contracts by status; what if we added the end time filter to that query as well? That should filter out a lot of results.

ljmcgann:
I also thought of that, but at the moment we do not have a filter that does that. I figured that work would be in some other refactoring pr

ljmcgann:
We do check the time before we go to request a lock however, so it doesn't decrease the number of db queries we make.

esi_leap/objects/contract.py
ljmcgann: My logic was that if an offer was not available, then it should have expired all of its contracts already.

tzumainn:
> I don't think you can use the decorator because we are using specific instances of the method from a class. If you go to the db, get a contract that is being expired, then go to the expire function and get blocked, that means that another process is currently performing the expire function. But the reference you grabbed out of the db will not have the changes reflecting that.

Looking at the contract code more closely, it does look like it's protected against being run multiple times consecutively, though. There may be a slight inefficiency to potentially having a contract expired twice - I guess if the lessee expires it right as it would naturally expire? - but dealing with that seems unneeded, since no harm would be done.

I think I'm more concerned about the cases where two people simultaneously try and create a contract for the same offer for overlapping times.

esi_leap/objects/contract.py
tzumainn: I'd remove this check, then; if the contract isn't expired, but the offer somehow has gotten into an expired state, I would think that we would want to expire that contract as fast as possible!

ljmcgann:
There is an overlapping time at creation for both contracts on an offer and offers on a resource issues and there is a contract creation at the same time an offer is being deleted issue. I am not sure if this work actually helps with either of those now though.

PR #54, opened 2020-08-18
Create a separate expire_contract

DanNiESh:
Create a separate expire_contract for ironic node, dummy node and test node. Undeploy the node when expiring an ironic node.

### DIFF
--- a/esi_leap/objects/contract.py
+++ b/esi_leap/objects/contract.py
@@ -86,7 +86,7 @@ class Contract(base.ESILEAPObject):
 
         resource = o.resource_object()
         if resource.get_contract_uuid() == self.uuid:
-            resource.set_contract(None)
+            resource.expire_contract(self)
 
         self.status = statuses.CANCELLED
         self.save(None)
@@ -117,7 +117,7 @@ class Contract(base.ESILEAPObject):
 
         resource = o.resource_object()
         if resource.get_contract_uuid() == self.uuid:
-            resource.set_contract(None)
+            resource.expire_contract(self)
 
         # expire contract
         self.status = statuses.EXPIRED
--- a/esi_leap/resource_objects/ironic_node.py
+++ b/esi_leap/resource_objects/ironic_node.py
@@ -59,30 +59,38 @@ class IronicNode(object):
 
     def set_contract(self, contract):
         patches = []
-        if contract is None:
-            if self.get_contract_uuid():
-                patches.append({
-                    "op": "remove",
-                    "path": "/properties/contract_uuid",
-                })
-            if self.get_project_id():
-                patches.append({
-                    "op": "remove",
-                    "path": "/lessee",
-                })
-        else:
+        patches.append({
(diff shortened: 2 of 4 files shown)


esi_leap/resource_objects/ironic_node.py
```diff
@@ -59,30 +59,36 @@ def get_node_config(self):
...
+        patches.append({
+            "op": "add",
+            "path": "/lessee",
+            "value": contract.project_id,
+        })
+        if len(patches) > 0:
+            get_ironic_client().node.update(self._uuid, patches)
+
+    def expire_contract(self, contract):
+        patches = []
```
tzumainn, line 76: Now that we can expire a specific contract, it may make sense to add something like:
```
if self.get_contract_uuid() != contract.uuid:
   return
```
This way there's no chance of accidentally expiring a node whose contract as already been assigned to a new contract.

esi_leap/resource_objects/ironic_node.py
```diff
@@ -59,30 +59,36 @@ def get_node_config(self):
...
-                "op": "add",
+                "op": "remove",
                 "path": "/lessee",
-                "value": contract.project_id,
             })
         if len(patches) > 0:
             get_ironic_client().node.update(self._uuid, patches)
+        state = get_ironic_client().node.get(self._uuid).provision_state
+        if state == "active":
+            get_ironic_client().node.set_provision_state(self._uuid, "deleted")
```
tzumainn, line 91: Would setting the provision_state to 'clean' work as well?

esi_leap/resource_objects/ironic_node.py
DanNiESh: Got it!

esi_leap/resource_objects/ironic_node.py
DanNiESh: I can try if it will work. Usually the undeployment stages are like active->deleting->cleaning->available. The doc says 'deleted' is used for a node transiting from 'active' to ' available'

tzumainn: Ah, in that case, leave it as deleted. Thanks!

esi_leap/objects/contract.py
```diff
@@ -117,7 +117,7 @@ def expire(self, context=None):
 
         resource = o.resource_object()
         if resource.get_contract_uuid() == self.uuid:
-            resource.set_contract(None)
+            resource.expire_contract(self)
```
tzumainn, line 120: Oh, I forgot - I think the 'cancel' method here also calls set_contract(None); can you update that to call expire_contract as well?

DanNiESh: Fixed it. Is there any command line corresponding with canceling a contract? I only see 'lease contract delete'

tzumainn: It's the delete call; we may have to rethink some of these CLI names at some point.

PR #56, opened 2020-08-19
Bug Fix on offer deletion and manager

ljmcgann:
Two bug fixes. One which prevented an owner from
deleting offers by uuid, and one where the manager could
not properly retrieve contracts to fulfill.

### DIFF
--- a/esi_leap/api/controllers/v1/utils.py
+++ b/esi_leap/api/controllers/v1/utils.py
@@ -27,7 +27,9 @@ def get_offer_authorized(uuid_or_name, cdict, status_filter=None):
 
         if not status_filter or o.status == status_filter:
             try:
-                policy.authorize('esi_leap:offer:offer_admin', cdict, cdict)
+                if o.project_id != cdict['project_id']:
+                    policy.authorize('esi_leap:offer:offer_admin',
+                                     cdict, cdict)
                 offer_objs.append(o)
             except oslo_policy.PolicyNotAuthorized:
                 pass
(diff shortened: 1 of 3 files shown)


esi_leap/api/controllers/v1/utils.py
```diff
@@ -27,7 +27,9 @@ def get_offer_authorized(uuid_or_name, cdict, status_filter=None):
 
         if not status_filter or o.status == status_filter:
             try:
-                policy.authorize('esi_leap:offer:offer_admin', cdict, cdict)
+                if o.project_id != cdict['project_id']:
```
tzumainn, line 30: Can you add tests that cover this being true and this being false?

esi_leap/api/controllers/v1/utils.py
```diff
@@ -65,8 +67,11 @@ def verify_resource_permission(cdict, offer_dict):
 def get_offer(uuid_or_name, status_filter=None):
     if uuidutils.is_uuid_like(uuid_or_name):
         o = offer.Offer.get(uuid_or_name)
-        if o.status == status_filter:
+        if not status_filter or o.status == status_filter:
             return o
+        else:
+            raise exception.OfferNotFound(
+                offer_uuid=uuid_or_name)
```
tzumainn, line 74: And tests that cover the various possibilities here (including the exception)?

esi_leap/api/controllers/v1/utils.py
ljmcgann: done.

esi_leap/api/controllers/v1/utils.py
ljmcgann: done

Issue #57, opened 2020-08-19
creating a contract with end-time earlier than start-time will raise issue


DanNiESh:
When creating a contract with an end-time, with or without a start-time, if the end-time is set to be earlier than current time (or start-time), the contract still gets created but raises errors in esi-leap-manager. And this contract cannot get deleted by cli.

PR #59, opened 2020-08-19
Bugfix on issue https://github.com/CCI-MOC/esi-leap/issues/57

ljmcgann:
Fixes the bug addressed in
https://github.com/CCI-MOC/esi-leap/issues/57


ljmcgann:
fixed

tzumainn:
Thanks!

PR #60, opened 2020-08-19
Reporting documentation

ljmcgann:
Added documentation generating reports

### DIFF
--- a/docs/esi-leap-reporting.md
+++ b/docs/esi-leap-reporting.md
@@ -1,0 +1,33 @@
+### Generating a report
+
+Esi-Leap along with the python-esi-leap client allow the user to easily generate reports on usage
+of a resource over time.
+
+To generate a csv file with data on all lessees' usage on different nodes:
+```
+openstack lease contract list --all --long -f csv > report.csv
+```
+
+The Esi-Leap api allows users to retrieve data based on certain parameters. 
+
+For example, it is possible to generate a report based on a lessee's openstack project_id. To get a list of all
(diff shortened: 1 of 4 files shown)


docs/esi-leap-reporting.md
```diff
@@ -0,0 +1,31 @@
...
+of a resource over time.
+
+To generate a csv file with data on all usage on different nodes:
+```
+openstack lease contract list --all --long -f csv > report.csv
+```
+
+The Esi-Leap api allows users to retrieve data based on certain parameters. 
+
+For example, it is possible to generate a report based on a user's openstack project_id. To get a list of all
```
tzumainn, line 13: This essentially views a lessee's usage, correct? If so, can we say "based on a lessee's openstack project_id"?

docs/esi-leap-reporting.md
```diff
@@ -0,0 +1,31 @@
...
+
+```
+openstack lease contract list --owner <project_id> --all --long -f csv > report.csv
+```
+
+To generate a report based on usage over a particular date range, run the command:
+
+```
+opensack lease contract list --time-range <datetime string> <datetime string> --all --long -f csv > report.csv
+```
```
tzumainn, line 31: Can you add some text explaining that multiple options can be combined together?

docs/esi-leap-reporting.md
ljmcgann: fixed

docs/esi-leap-reporting.md
ljmcgann: done

PR #65, opened 2021-04-13
Add owner field for leases

tzumainn:
This change allows for easier reporting by node owner; and
accommodates the case where an owner may want to create a lease
without an offer.


ljmcgann:
What does creating a contract, err lease, without an offer mean? What is the lease on?

tzumainn:
> What does creating a contract, err lease, without an offer mean? What is the lease on?

It's not yet supported by the API, but the idea is that a node owner can grant a lease directly to a project, without creating an offer. So eventually, a node owner will be able to create a lease on behalf of a lessee.

ljmcgann:
I see. So offers were only ever a structure for determining whether a lease could be created, and maybe we want to skip that step sometimes. Makes sense to me

esi_leap/db/sqlalchemy/api.py
```diff
@@ -287,7 +287,6 @@ def lease_get_all(filters):
 
     start = filters.pop('start_time', None)
     end = filters.pop('end_time', None)
-    owner = filters.pop('owner', None)
```
DanNiESh, line 290: I don't understand why owner has been removed in filters here. Looks like the Lease data model has a new lease_owner_id_idx index but it's not been used?

ljmcgann:
looks good to me

DanNiESh (commented):
Looks good overall! I left a question in one place.

esi_leap/db/sqlalchemy/api.py
tzumainn: What this code does is actually remove the special filters - for example start and end - and use special processing for those specific filters that you can see in the code after this line. The fields that are *not* popped are used as a simple equality filter. This line is removed because now we no longer have to use special filtering processing (joining with an offer) for an owner.

ljmcgann: my brilliant code at work

DanNiESh: I see!

DanNiESh (approved):
Thanks!

PR #67, opened 2021-04-22
Add default resource type config

tzumainn:
This change adds a default resource type in the configuration. If
unset, it defaults to 'ironic_node'

### DIFF
--- a/esi_leap/tests/api/controllers/v1/test_offer.py
+++ b/esi_leap/tests/api/controllers/v1/test_offer.py
@@ -52,6 +52,16 @@ class TestOffersController(test_api_base.APITestCase):
             end_time=start + datetime.timedelta(days=100),
             project_id=self.context.project_id,
         )
+        self.test_offer_drt = offer.Offer(
+            resource_type='ironic_node',
+            resource_uuid=uuidutils.generate_uuid(),
+            name="test_offer",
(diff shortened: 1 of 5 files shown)


esi_leap/tests/api/controllers/v1/test_offer.py
```diff
@@ -105,6 +115,41 @@ def test_post(self, mock_aoa, mock_create, mock_cra, mock_generate_uuid):
...
+        }
+
+        request = self.post_json('/offers', data)
+
+        data['project_id'] = self.context.project_id
+        data['uuid'] = self.test_offer_default_resource_type.uuid
+        data['status'] = statuses.AVAILABLE
+        data['resource_type'] = 'ironic_node'
+        data['resource_uuid'] = \
+            self.test_offer_default_resource_type.resource_uuid
```
DanNiESh, line 142: data['resource_uuid'] before sending the post request is self.test_offer.resource_uuid, and now it's set as self.test_offer_default_resource_type.uuid. I don't get it, though it won't effect the test.

tzumainn: Good point! I've updated the test to be more consistent

DanNiESh (approved):

PR #68, opened 2021-04-23
Update default lease get all to include owned leases

tzumainn:
Previously, getting all leases defaulted to only showing the user
leases that they were lessees for. This change adds leases that
the user is the owner for.

### DIFF
--- a/esi_leap/api/controllers/v1/lease.py
+++ b/esi_leap/api/controllers/v1/lease.py
@@ -77,7 +77,7 @@ class LeasesController(rest.RestController):
                          datetime.datetime, datetime.datetime, wtypes.text,
                          wtypes.text, wtypes.text, wtypes.text)
     def get_all(self, project_id=None, start_time=None, end_time=None,
-                status=None, offer_uuid=None, view=None, owner=None):
+                status=None, offer_uuid=None, view=None, owner_id=None):
         request = pecan.request.context
         cdict = request.to_policy_values()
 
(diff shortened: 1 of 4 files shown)


esi_leap/api/controllers/v1/lease.py
```diff
@@ -155,26 +155,27 @@ def _lease_get_all_authorize_filters(cdict,
...
                 possible_filters['project_id'] = project_id
             else:
                 if project_id is None:
                     project_id = cdict['project_id']
-                elif project_id != cdict['project_id']:
-                    policy.authorize('esi_leap:lease:lease_admin',
-                                     cdict, cdict)
-
-                possible_filters['project_id'] = project_id
+                    possible_filters['project_or_owner_id'] = project_id
```
ljmcgann, line 173: is github being weird or is this tabbed in when it shouldnt be

ljmcgann (approved):
looks good to me

esi_leap/api/controllers/v1/lease.py
tzumainn: Ah, the indent is correct - we only set 'project_or_owner_id' if the user has decided *not* to filter by project_id. Thanks for the review!

ljmcgann: oh i was being dumb and read the old line with the new else statement and was confused. All else considered yes it looks fine.

ljmcgann (approved):

PR #69, opened 2021-04-27
Add lessee_id to offers

tzumainn:
If an offer has a lessee_id, then it can only be claimed by the
project matching that lessee_id or subprojects of that project.

### DIFF
--- a/esi_leap/common/keystone.py
+++ b/esi_leap/common/keystone.py
@@ -1,0 +1,43 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from keystoneauth1 import loading as ks_loading
+from keystoneclient import client as keystone_client
+
+import esi_leap.conf
+
+
+CONF = esi_leap.conf.CONF
+_cached_keystone_client = None
+
+
+def get_keystone_client():
+    global _cached_keystone_client
+    if _cached_keystone_client is not None:
+        return _cached_keystone_client
+
+    auth_plugin = ks_loading.load_auth_from_conf_options(CONF, 'keystone')
+    sess = ks_loading.load_session_from_conf_options(CONF, 'keystone',
+                                                     auth=auth_plugin)
+    cli = keystone_client.Client(session=sess)
+    _cached_keystone_client = cli
+
+    return cli
+
+
+def get_parent_project_id_tree(project_id):
+    project = get_keystone_client().projects.get(project_id)
+    project_ids = [project.id]
+    while project.parent_id is not None:
+        project = get_keystone_client().projects.get(project.parent_id)
+        project_ids.append(project.id)
+    return project_ids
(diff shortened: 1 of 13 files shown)


esi_leap/common/keystone.py
```diff
@@ -0,0 +1,43 @@
...
+    auth_plugin = ks_loading.load_auth_from_conf_options(CONF, 'keystone')
+    sess = ks_loading.load_session_from_conf_options(CONF, 'keystone',
+                                                     auth=auth_plugin)
+    cli = keystone_client.Client(session=sess)
+    _cached_keystone_client = cli
+
+    return cli
+
+
+def get_parent_project_id_tree(project_id):
```
ljmcgann, line 37: i have never seen recursion in industry code before. Is there any reason we are using recursion here instead of a for loop?

DanNiESh, line 37: What is a parent project and why do we need to collect the parent project tree here?

tzumainn: The model here is this: there are two separate projects that represent two different organizations - A and B. Each organization has multiple projects in their domain. A and B are node owners who may have temporary access to a given node.

This code is intended to solve the use case where A wants to offer up a node they own, but only to its own subprojects. In order to do so, they specify the lessee_id to be themselves, and this code will hide that offer from any project other than projects in their own tree.

esi_leap/db/sqlalchemy/models.py
```diff
@@ -50,6 +50,7 @@ class Offer(Base):
     uuid = Column(String(36), nullable=False, unique=True)
     name = Column(String(35), nullable=True, unique=False)
     project_id = Column(String(255), nullable=False)
+    lessee_id = Column(String(255))
```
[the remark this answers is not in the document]
tzumainn: Yep! I'll add that in, although it's true by default.

esi_leap/common/keystone.py
tzumainn: Well, a parent project could have its own parents, etc. But you're probably right; I'll re-write this as a while loop.

tzumainn:
Thanks for the reviews! I've updated the code based on your comments - let me know if there are any questions!

esi_leap/db/sqlalchemy/api.py
```diff
@@ -112,14 +114,20 @@ def offer_get_all(filters):
...
-
     a_start = filters.pop('available_start_time', None)
     a_end = filters.pop('available_end_time', None)
 
     query = query.filter_by(**filters)
 
+    if lessee_id:
+        lessee_id_list = keystone.get_parent_project_id_tree(lessee_id)
+        query = query.filter(or_(models.Offer.project_id == lessee_id,
+                                 models.Offer.lessee_id.__eq__(None),
```
DanNiESh, line 128: I'm curious why lessee_id can be none when filtering by lessee_id?

tzumainn: Ah, that's because if the offer has no lessee, that means that it's not limited, and anyone can claim it.

DanNiESh: Thanks! I see.

esi_leap/db/sqlalchemy/models.py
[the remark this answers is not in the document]
DanNiESh: Solved.

DanNiESh (approved):
Looks good to me!

esi_leap/common/keystone.py
ljmcgann: is lessee_id the best name for this? It wasnt immediately obvious to me that that would be the field name.

ljmcgann: Like something that implies its a family of ids rather than just one specific person. Just a random nit I have

ljmcgann:
looks good to me

ljmcgann (approved):

PR #70, opened 2021-05-06
Add owner change to leasing system

tzumainn:
An owner change allows a user - by default, only admins - to
specify a length of time during which a resource object will
change ownership. The manager will take care of activating and
reverting that change.

### DIFF
--- a/esi_leap/api/controllers/v1/utils.py
+++ b/esi_leap/api/controllers/v1/utils.py
@@ -56,10 +56,11 @@ def get_offer_authorized(uuid_or_name, cdict, status_filter=None):
     return offer_objs[0]
 
 
-def check_resource_admin(cdict, resource_type, resource_uuid, project_id):
+def check_resource_admin(cdict, resource_type, resource_uuid, project_id,
+                         start_time, end_time):
     resource = ro_factory.ResourceObjectFactory.get_resource_object(
         resource_type, resource_uuid)
-    if not resource.is_resource_admin(project_id):
+    if not resource.check_admin(project_id, start_time, end_time):
         policy.authorize('esi_leap:offer:offer_admin', cdict, cdict)
 
 
--- a/esi_leap/db/api.py
+++ b/esi_leap/db/api.py
@@ -146,3 +146,12 @@ def lease_destroy(lease_uuid):
 def resource_verify_availability(r_type, r_uuid, start, end):
     return IMPL.resource_verify_availability(
         r_type, r_uuid, start, end)
+
+
+def resource_check_admin(resource_type, resource_uuid,
+                         start_time, end_time,
+                         default_admin_project_id, project_id):
+    return IMPL.resource_check_admin(
+        resource_type, resource_uuid,
+        start_time, end_time,
+        default_admin_project_id, project_id)
--- a/esi_leap/db/sqlalchemy/api.py
+++ b/esi_leap/db/sqlalchemy/api.py
@@ -117,6 +117,7 @@ def offer_get_all(filters):
     lessee_id = filters.pop('lessee_id', None)
     start = filters.pop('start_time', None)
     end = filters.pop('end_time', None)
+    time_filter_type = filters.pop('time_filter_type', None)
     a_start = filters.pop('available_start_time', None)
     a_end = filters.pop('available_end_time', None)
 
@@ -129,8 +130,15 @@ def offer_get_all(filters):
                                  models.Offer.lessee_id.in_(lessee_id_list)))
 
     if start and end:
-        query = query.filter((start >= models.Offer.start_time) &
-                             (end <= models.Offer.end_time))
+        if time_filter_type == 'within':
+            query = query.filter(((start <= models.Offer.start_time) &
+                                  (end >= models.Offer.start_time)) |
+
+                                 ((start <= models.Offer.end_time) &
+                                  (end >= models.Offer.end_time)))
+        else:
+            query = query.filter((start >= models.Offer.start_time) &
+                                 (end <= models.Offer.end_time))
 
     if a_start and a_end:
         for o in query:
@@ -186,14 +194,14 @@ def offer_verify_availability(offer_ref, start, end):
                )
 
     conflict = leases.filter((
-        (start >= models.Lease.start_time) &
-        (start < models.Lease.end_time) |
+        ((start >= models.Lease.start_time) &
+         (start < models.Lease.end_time)) |
 
-        (end > models.Lease.start_time) &
-        (end <= models.Lease.end_time) |
+        ((end > models.Lease.start_time) &
+         (end <= models.Lease.end_time)) |
 
-        (start <= models.Lease.start_time) &
-        (end >= models.Lease.end_time)
+        ((start <= models.Lease.start_time) &
+         (end >= models.Lease.end_time))
     )).first()
 
     if conflict:
@@ -268,6 +276,7 @@ def lease_get_all(filters):
 
     start = filters.pop('start_time', None)
     end = filters.pop('end_time', None)
+    time_filter_type = filters.pop('time_filter_type', None)
     status = filters.pop('status', None)
     project_or_owner_id = filters.pop('project_or_owner_id', None)
 
@@ -277,8 +286,15 @@ def lease_get_all(filters):
         query = query.filter((models.Lease.status.in_(status)))
 
     if start and end:
-        query = query.filter((start >= models.Lease.start_time) &
-                             (end <= models.Lease.end_time))
+        if time_filter_type == 'within':
+            query = query.filter(((start <= models.Lease.start_time) &
+                                  (end >= models.Lease.start_time)) |
+
+                                 ((start <= models.Lease.end_time) &
+                                  (end >= models.Lease.end_time)))
+        else:
+            query = query.filter((start >= models.Lease.start_time) &
+                                 (end <= models.Lease.end_time))
 
     if project_or_owner_id:
         query = query.filter(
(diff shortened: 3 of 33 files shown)


esi_leap/db/sqlalchemy/models.py
```diff
@@ -91,3 +91,26 @@ class Lease(Base):
...
+
+    id = Column(Integer, primary_key=True, nullable=False, autoincrement=True)
+    uuid = Column(String(36), nullable=False, unique=True)
+    from_owner_id = Column(String(255), nullable=False)
+    to_owner_id = Column(String(255), nullable=False)
+    resource_type = Column(String(36), nullable=False)
+    resource_uuid = Column(String(36), nullable=False)
+    start_time = Column(DateTime)
+    end_time = Column(DateTime)
+    fulfill_time = Column(DateTime)
```
DanNiESh, line 114: Does fulfill time refer to the ownership changing time?

esi_leap/objects/owner_change.py
```diff
@@ -0,0 +1,185 @@
...
+        db_owner_change = cls.dbapi.owner_change_get_by_uuid(owner_change_uuid)
+        if db_owner_change:
+            return cls._from_db_object(context, cls(), db_owner_change)
+
+    @classmethod
+    def get_all(cls, filters, context=None):
+        db_owner_changes = cls.dbapi.owner_change_get_all(filters)
+        return cls._from_db_object_list(context, db_owner_changes)
+
+    def create(self, context=None):
```
DanNiESh, line 62: Do we need to check if from_owner_id != to_owner_id?

esi_leap/db/sqlalchemy/models.py
tzumainn: Yep! Same nomenclature as for leases.

esi_leap/objects/owner_change.py
tzumainn: Technically nothing bad would happen, but it's probably worth checking. I've added this check (and a corresponding test) in the API controller.

esi_leap/db/sqlalchemy/api.py
```diff
@@ -334,7 +350,93 @@ def lease_destroy(lease_uuid):
...
+        if not owner_change_ref:
+            raise exception.OwnerChangeNotFound(
+                owner_change_uuid=owner_change_uuid)
+        query.delete()
+        session.flush()
+
+
+# Resources
+def resource_verify_availability(r_type, r_uuid, start, end,
+                                 is_owner_change=False):
```
ljmcgann, line 439: is it necessary to do these offer and lease checks when verifying a time for an owner_change object?

ljmcgann: this is the one part that is a bit tricky to wrap my head around

esi_leap/db/api.py
```diff
@@ -146,3 +146,12 @@ def lease_destroy(lease_uuid):
 def resource_verify_availability(r_type, r_uuid, start, end):
     return IMPL.resource_verify_availability(
         r_type, r_uuid, start, end)
+
+
+def resource_check_admin(resource_type, resource_uuid,
+                         start_time, end_time,
+                         default_admin_project_id, project_id):
+    return IMPL.resource_verify_availability(
```
DanNiESh, line 154: Why this function isn't mapping to resource_check_admin() in db/sqlalchemy/api.py?

esi_leap/db/sqlalchemy/api.py
```diff
@@ -370,17 +472,94 @@ def resource_verify_availability(r_type, r_uuid, start, end):
...
+        'end_time': end_time,
+        'status': [statuses.CREATED, statuses.ACTIVE]
+    }
+    ocs = owner_change_get_all(filters)
+
+    if ocs.count() > 1:
+        # shouldn't happen, but...
+        return False
+    if ocs.count() == 1:
+        return project_id == ocs[0].to_owner_id
```
DanNiESh, line 563: I don't understand here... Does resource_check_admin() check if the given project is the admin project?

esi_leap/db/sqlalchemy/api.py
tzumainn: Yeah, this one was extremely difficult for me to wrap my head around as well. My reasoning here is that if owner A creates a lease or an offer for a time period, it's illogical to allow them to move ownership of the node to owner B, as it's unclear whether owner B should have the right to cancel those leases/offers, or even to see them.

esi_leap/db/api.py
tzumainn: Whoops - nice catch. Fixed!

esi_leap/db/sqlalchemy/api.py
DanNiESh: I see! Thanks!

esi_leap/objects/owner_change.py
```diff
@@ -0,0 +1,185 @@
...
+            self._from_db_object(context, self, db_owner_change)
+
+        _create_owner_change()
+
+    def fulfill(self, context=None):
+        @utils.synchronized(utils.get_resource_lock_name(self.resource_type,
+                                                         self.resource_uuid),
+                            external=True)
+        def _fulfill_owner_change():
+            LOG.info("Setting owner of %s %s to %s",
```
DanNiESh, line 94: I think it's better to contain from_owner_id in logs.

esi_leap/objects/owner_change.py
```diff
@@ -0,0 +1,185 @@
...
+                     self.resource_type, self.resource_uuid, self.to_owner_id)
+
+            resource = self.resource_object()
+            resource.set_owner(self.to_owner_id)
+
+            # activate lease
```
DanNiESh, line 100: typo: active owner_change

esi_leap/objects/owner_change.py
```diff
@@ -0,0 +1,185 @@
...
+            self.status = statuses.ACTIVE
+            self.fulfill_time = datetime.datetime.now()
+            self.save(context)
+
+        _fulfill_owner_change()
+
+    def cancel(self):
```
DanNiESh, line 107: Is it proper to cancel an owner change when a lease is in the active status? Also, for unclaimed offers, when an owner change is cancelled, can we simply update the offer with the original owner?

esi_leap/objects/owner_change.py
```diff
@@ -0,0 +1,185 @@
...
+            resource = self.resource_object()
+            resource.set_owner(self.from_owner_id)
+
+            self.status = statuses.CANCELLED
+            self.expire_time = datetime.datetime.now()
+            self.save(None)
+
+        _cancel_owner_change()
+
+    def expire(self, context=None):
```
DanNiESh, line 131: I guess in theory, for a resource, no offer or lease will be created during an owner change?

esi_leap/objects/owner_change.py
tzumainn: I'll add it in!

esi_leap/objects/owner_change.py
tzumainn: Whoops - fixed!

esi_leap/objects/owner_change.py
DanNiESh: I have this confusion because creating an offer or a lease calls verify_availability() in resource_objects, and this function checks if there're time conflicts with other offers, leases, and owner changes on this resource and raises exceptions when conflicts happen. Not sure if I understand it correctly.

tzumainn: Ah, yes! If you look at `resource_verify_availability` in `db/sqlalchemy/api.py`, it takes a flag: `is_owner_change`. Leases and offers call verify_availability with that flag set to false, meaning that they specifically do *not* check against `owner_changes`; instead they would have used `check_resource_admin` earlier to verify that the lease/offer created is compatible with whoever the admin is at the time, as per the owner change.

DanNiESh: oh, I see! It makes sense to me now.

esi_leap/api/controllers/v1/utils.py
```diff
@@ -56,10 +56,11 @@ def get_offer_authorized(uuid_or_name, cdict, status_filter=None):
...
 
 
-def check_resource_admin(cdict, resource_type, resource_uuid, project_id):
+def check_resource_admin(cdict, resource_type, resource_uuid, project_id,
+                         start_time, end_time):
     resource = ro_factory.ResourceObjectFactory.get_resource_object(
         resource_type, resource_uuid)
-    if not resource.is_resource_admin(project_id):
+    if not resource.check_admin(project_id, start_time, end_time):
         policy.authorize('esi_leap:offer:offer_admin', cdict, cdict)
```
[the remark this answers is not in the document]
tzumainn: Well, this check only verifies that the project is an admin for the passed-in `start_time`/`end_time`. If a project passes this check and creates an offer/lease, and then someone else tries to create an owner change during that time period, the owner change creation will fail due to the existence of that offer/lease.

esi_leap/tests/objects/test_owner_change.py
```diff
@@ -0,0 +1,265 @@
...
+            'updated_at': None
+        }
+
+        oc = owner_change.OwnerChange(self.context, **bad_oc_data)
+
+        self.assertRaises(exception.InvalidTimeRange, oc.create)
+
+    @mock.patch('esi_leap.db.sqlalchemy.api.resource_verify_availability')
+    @mock.patch('esi_leap.db.sqlalchemy.api.owner_change_create')
+    def test_create_concurrent(self, mock_occ, mock_rva):
```
DanNiESh, line 123: I don't think this test proves create_owner_change is thread safe cus it's not a deterministic test.

esi_leap/tests/api/controllers/v1/test_owner_change.py
```diff
@@ -0,0 +1,409 @@
...
+            'esi_leap:owner_change:owner_change_admin',
+            self.context.to_policy_values(),
+            self.context.to_policy_values()
+        )
+        mock_get_all.assert_called_once_with(expected_filters, self.context)
+        self.assertEqual(expected_resp, response)
+
+    @mock.patch('esi_leap.objects.owner_change.OwnerChange.get_all')
+    @mock.patch('esi_leap.common.policy.authorize')
+    def test_get_all_invalid_time(self, mock_authorize, mock_get_all):
```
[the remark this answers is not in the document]
tzumainn: Added!

esi_leap/tests/objects/test_owner_change.py
tzumainn: It seems deterministic to me... ? Perhaps it's worth looking into later, but I'm copying the pattern from test_lease.py and test_offer.py. I'm not sure there's a better way to test this.

DanNiESh: My thought is that assuming the function is prone to race conditions, the test may or may not detect this issue each time it's executed. But I'm not sure... It's may worth looking into later.

tzumainn: I'll create an issue for looking at the concurrency tests!

DanNiESh (approved):
Looks good to me!

PR #72, opened 2021-05-20
Update resource admin db api

tzumainn:
Previously, the resource admin db api would take in various project
IDs and check for a match within the db api. This prevents the code
from distinguishing between the cases where the check fails because
of a time conflict, and when the check fails because the passed in
project IDs do not match. That, in turn, prevents us from effectively
checking the admin case, as that case should only be checked if there
is no time conflict.

This change updates the resource admin db api to simply return the
project ID as determined by the owner change, allowing us to more
effectively route our way through various alternatives at the API
level.


DanNiESh (approved):
Looks good.

tzumainn:
No longer needed as owner changes are going away.

PR #73, opened 2021-06-11
Allow nodes and projects to be specified by name in the REST API

tzumainn:
Nodes and projects are identified by UUID in the database, but
allowing users to specify them by UUID *or* name when using the
REST API will be very convenient.

### DIFF
--- a/esi_leap/tests/api/controllers/v1/test_lease.py
+++ b/esi_leap/tests/api/controllers/v1/test_lease.py
@@ -21,6 +21,8 @@ import testtools
 from esi_leap.api.controllers.v1.lease import LeasesController
 from esi_leap.common import exception
 from esi_leap.objects import lease as lease_obj
+from esi_leap.resource_objects.ironic_node import IronicNode
+from esi_leap.resource_objects.test_node import TestNode
 from esi_leap.tests.api import base as test_api_base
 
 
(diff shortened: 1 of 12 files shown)


tzumainn:
owner changes are skipped, as we are planning on removing that functionality.

esi_leap/tests/api/controllers/v1/test_lease.py
```diff
@@ -109,6 +125,121 @@ def test_post_default_resource_type(self, mock_create, mock_cra,
...
+                                           project_id=None,
+                                           start_time=None,
+                                           end_time=None,
+                                           status=None,
+                                           offer_uuid=None,
+                                           view=None,
+                                           owner_id=None,
+                                           resource_type=None,
+                                           resource_uuid=None)
+        mock_get_all.assert_called_once
```
DanNiESh, line 144: assert_called_once(), just my personal preference to make it consistent with other tests. Here and other assert_called_once in this file.

esi_leap/tests/api/controllers/v1/test_offer.py
```diff
@@ -84,12 +97,16 @@ def test_one(self):
...
 
+    @mock.patch('esi_leap.api.controllers.v1.utils.ro_factory.'
+                'ResourceObjectFactory.get_resource_object')
     @mock.patch('oslo_utils.uuidutils.generate_uuid')
     @mock.patch('esi_leap.api.controllers.v1.utils.check_resource_admin')
     @mock.patch('esi_leap.objects.offer.Offer.create')
     @mock.patch('esi_leap.api.controllers.v1.offer.' +
                 'OffersController._add_offer_availabilities')
-    def test_post(self, mock_aoa, mock_create, mock_cra, mock_generate_uuid):
+    def test_post(self, mock_aoa, mock_create, mock_cra, mock_generate_uuid,
```
DanNiESh, line 107: I wonder do we need tests on post/get with project_name? Seems like test_keystone is enough but I'm not sure.

tzumainn: Good question! For unit testing purposes, it's enough to have test_keystone, as test_keystone ensures that the project-name-or-uuid functionality works, while test_offer just ensures that we call the keystone function.

esi_leap/tests/api/controllers/v1/test_lease.py
tzumainn: Fair point! Fixed.

tzumainn:
Thanks for the review! I think I've addressed the issues; let me know if there's anything else!

esi_leap/tests/api/controllers/v1/test_offer.py
DanNiESh: Sounds fair.

DanNiESh (approved):
Looks good to me. I don't see any problems.

PR #78, opened 2021-07-09
Catch PolicyNotAuthorized errors and replace with HTTPForbidden

tzumainn:
PolicyNotAuthorized is confusing to end-users; HTTPForbidden is
more understandable, and also more accurate.


DanNiESh (approved):
Looks good to me!

PR #79, opened 2021-07-15
Disallow duplicate lease and offer names

tzumainn:
Allowing duplicate names complicates the code to little benefit.
Cleaning up this code simplifies it tremendously, and leaves less
room for potential edge case failures.

### DIFF
--- a/esi_leap/api/controllers/v1/utils.py
+++ b/esi_leap/api/controllers/v1/utils.py
@@ -21,39 +21,16 @@ from esi_leap.objects import offer as offer_obj
 
 
 def get_offer_authorized(uuid_or_name, cdict, status_filter=None):
-    if uuidutils.is_uuid_like(uuid_or_name):
-        o = offer_obj.Offer.get(uuid_or_name)
-        offer_objs = []
+    offer = get_offer(uuid_or_name, status_filter)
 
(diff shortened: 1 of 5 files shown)


larsks:
> Allowing duplicate names complicates the code to little benefit.

Avoiding duplicates names within a project seems like a good idea, but avoiding duplicate names across projects is problematic. If I want to create a lease named `my-example-lease` and it fails because someone in another projects -- for which I have no visibility -- picked the same name is a poor user experience.

It wasn't clear to me which behavior is implemented by this change.

tzumainn:
Oh, that's a really good point; I completely missed that because ironic node names are unique - but the data model there wasn't created with multi-tenancy in mind. That being the case, this update may have little value.

tzumainn:
Thinking it about a little bit more, in the back of my mind I was thinking primarily about offers, which cross project boundaries. I think you're right about about the unique name constraint being unwanted for leases, but what do you think about offers?

larsks:
If it's just for a globally visible resource like offers, that makes more sense.

tzumainn (commented):
I'm going to close this for now until we figure out how people may actually want to use names. Auto-generated actually sounds somewhat attractive to me, but may as well wait for actual user requests, I think.

PR #82, opened 2021-07-15
Add resource list functionality

betaredex:
Set up a NodesController and an API path for getting all ironic nodes.
Right now, it only returns the names of the nodes, and the get_ironic_client method in esi_leap/common/ironic.py is copied verbatim from esi_leap/resource_objects/ironic_node.py. Ideally, we'd use the current user's credentials instead of the admin's.

### DIFF
--- a/esi_leap/api/controllers/v1/node.py
+++ b/esi_leap/api/controllers/v1/node.py
@@ -1,0 +1,52 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import pecan
+from pecan import rest
+import wsme
+from wsme import types as wtypes
+import wsmeext.pecan as wsme_pecan
+
+from esi_leap.api.controllers import base
+from esi_leap.api.controllers import types
+from esi_leap.common import ironic
+import esi_leap.conf
+
+CONF = esi_leap.conf.CONF
+
+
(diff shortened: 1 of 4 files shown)


esi_leap/api/controllers/v1/node.py
```diff
@@ -0,0 +1,52 @@
...
+
+class NodesController(rest.RestController):
+    
+    @wsme_pecan.wsexpose(NodeCollection)
+    def get_all(self):
+        context = pecan.request.context
+        client = ironic.get_ironic_client(context)
+        node_collection = NodeCollection()
+        nodes = client.node.list()
+        print(nodes)
```
tzumainn, line 49: This should probably go.

esi_leap/common/ironic.py
```diff
@@ -0,0 +1,42 @@
...
+#    under the License.
+
+from keystoneauth1 import loading as ks_loading
+from keystoneauth1 import service_token
+from keystoneauth1 import token_endpoint
+
+from ironicclient import client as ironic_client
+
+import esi_leap.conf
+from esi_leap.resource_objects import base
```
tzumainn, line 20: This can be removed.

esi_leap/tests/api/controllers/v1/test_node.py
```diff
@@ -0,0 +1,33 @@
...
+
+class TestNodesController(test_api_base.APITestCase):
+
+    def setUp(self):
+        super(TestNodesController, self).setUp()
+
+    @mock.patch.object(ironic, 'get_ironic_client', autospec=True)
+    def test_get_all(self, client_mock):
+        fake_node = FakeIronicNode()
+        client_mock.return_value.node.get.return_value = fake_node
```
tzumainn, line 30: I think you might want something like:
```
client_mock.return_value.node.list.return_value = [fake_node]
```
Since the function doesn't call node.get; it calls node.list

tzumainn (commented):
Not bad! A few things to fix up.

esi_leap/tests/api/controllers/v1/test_node.py
betaredex: Done

betaredex:
Fixed issues and added a check that the fake node is actually returned

tzumainn (approved):
Looks good - thanks!

PR #83, opened 2021-07-21
Optimize name queries for list commands

tzumainn:
Optimize name queries for list commands by grabbing a list of
all projects/nodes and searching the list (rather than doing
an API request per project/node).

### DIFF
--- a/esi_leap/api/controllers/v1/lease.py
+++ b/esi_leap/api/controllers/v1/lease.py
@@ -23,6 +23,7 @@ from esi_leap.api.controllers import base
 from esi_leap.api.controllers import types
 from esi_leap.api.controllers.v1 import utils
 from esi_leap.common import exception
+from esi_leap.common import ironic
 from esi_leap.common import keystone
 from esi_leap.common import statuses
 import esi_leap.conf
@@ -114,9 +115,13 @@ class LeasesController(rest.RestController):
         lease_collection = LeaseCollection()
         leases = lease_obj.Lease.get_all(filters, request)
         lease_collection.leases = []
-        for lease in leases:
-            lease_collection.leases.append(
-                Lease(**LeasesController._lease_get_dict_with_names(lease)))
+        if len(leases) > 0:
+            project_list = keystone.get_project_list()
+            node_list = ironic.get_node_list()
+            for lease in leases:
+                lease_collection.leases.append(
+                    Lease(**LeasesController._lease_get_dict_with_names(
+                        lease, project_list, node_list)))
         return lease_collection
 
     @wsme_pecan.wsexpose(Lease, body=Lease, status_code=http_client.CREATED)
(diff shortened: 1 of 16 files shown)


esi_leap/common/ironic.py
```diff
@@ -23,19 +23,40 @@
...
+    user_auth = None
+    if context:
+        endpoint = ks_loading.load_adapter_from_conf_options(
+            CONF, 'ironic', session=session, auth=service_auth).get_endpoint()
+        user_auth = service_token.ServiceTokenAuthWrapper(
+            user_auth=token_endpoint.Token(endpoint, context.auth_token),
+            service_auth=service_auth)
     sess = ks_loading.load_session_from_conf_options(
-        CONF, 'ironic', auth=user_auth)
+        CONF, 'ironic', auth=user_auth or service_auth)
```
DanNiESh, line 40: When will service_auth be used instead of user_auth?

esi_leap/resource_objects/dummy_node.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, uuid):
     def get_resource_uuid(self):
         return self._uuid
 
-    def get_resource_name(self):
+    def get_resource_name(self, resource_list=None):
         return "dummy-node-%s" % self._uuid
```
DanNiESh, line 35: So we don't implement resource_list for dummy nodes?

esi_leap/common/ironic.py
tzumainn: Good question! It should be used when user_auth is None - in other words, when a user context isn't passed in. I was unfamiliar with this pattern, but Ironic uses it: https://opendev.org/openstack/ironic/src/branch/master/ironic/common/neutron.py#L76

esi_leap/resource_objects/dummy_node.py
tzumainn: Nope. We need to query Ironic for ironic_nodes, but there's no need to query a name for dummy/test nodes. I was actually torn between putting an `if` statement in the controller to retrieve the name in this special way only if it's an ironic node, and doing it this way which allows the controller code to be slightly cleaner. I don't have a strong opinion one way or the other though!

esi_leap/common/ironic.py
DanNiESh: Got it!

esi_leap/api/controllers/v1/lease.py
```diff
@@ -114,9 +115,13 @@ def get_all(self, project_id=None, start_time=None, end_time=None,
         lease_collection = LeaseCollection()
         leases = lease_obj.Lease.get_all(filters, request)
         lease_collection.leases = []
-        for lease in leases:
-            lease_collection.leases.append(
-                Lease(**LeasesController._lease_get_dict_with_names(lease)))
+        if len(leases) > 0:
+            project_list = keystone.get_project_list()
+            node_list = ironic.get_node_list()
```
DanNiESh, line 120: Does context parameter need to be passed to get_node_list()?

esi_leap/api/controllers/v1/offer.py
```diff
@@ -166,10 +167,15 @@ def get_all(self, project_id=None, resource_type=None,
 
         offer_collection = OfferCollection()
         offers = offer_obj.Offer.get_all(filters, request)
+        offer_collection.offers = []
+        if len(offers) > 0:
+            project_list = keystone.get_project_list()
+            node_list = ironic.get_node_list()
```
DanNiESh, line 173: same question

esi_leap/api/controllers/v1/lease.py
tzumainn: Nope! The context parameter is the user context, which would limit the retrieved nodes to the one that the user can see in Ironic. But if we want them to see the names of nodes that are on offer, then we use no context (which results in an ironic client that uses the admin context), so that they can see all of those nodes.

esi_leap/api/controllers/v1/offer.py
tzumainn: same answer :)

esi_leap/api/controllers/v1/lease.py
DanNiESh: oh, I see! Thanks!

DanNiESh (approved):
LGTM. I don't see any problems.

PR #84, opened 2021-07-23
Use is_lease_* and is_offer_* policy rules

tzumainn:
This change moves us more in line with OpenStack standards, and
also simplifies some of our controller logic by falling back to
oslo policy.

### DIFF
--- a/esi_leap/api/controllers/v1/lease.py
+++ b/esi_leap/api/controllers/v1/lease.py
@@ -74,10 +74,9 @@ class LeasesController(rest.RestController):
     @wsme_pecan.wsexpose(Lease, wtypes.text)
     def get_one(self, lease_id):
         request = pecan.request.context
-        cdict = request.to_policy_values()
-        utils.policy_authorize('esi_leap:lease:get', cdict)
 
-        lease = utils.get_lease_authorized(lease_id, cdict)
+        lease = utils.check_lease_policy_and_retrieve(
+            request, 'esi_leap:lease:get', lease_id)
 
         return Lease(**LeasesController._lease_get_dict_with_names(lease))
 
@@ -128,7 +127,7 @@ class LeasesController(rest.RestController):
     def post(self, new_lease):
         request = pecan.request.context
         cdict = request.to_policy_values()
-        utils.policy_authorize('esi_leap:lease:create', cdict)
+        utils.policy_authorize('esi_leap:lease:create', cdict, cdict)
 
         lease_dict = new_lease.to_dict()
         lease_dict['owner_id'] = request.project_id
@@ -174,13 +173,12 @@ class LeasesController(rest.RestController):
     @wsme_pecan.wsexpose(Lease, wtypes.text)
     def delete(self, lease_id):
         request = pecan.request.context
-        cdict = request.to_policy_values()
-        utils.policy_authorize('esi_leap:lease:delete', cdict)
 
-        permitted = utils.get_lease_authorized(
-            lease_id, cdict, [statuses.CREATED, statuses.ACTIVE])
+        lease = utils.check_lease_policy_and_retrieve(
+            request, 'esi_leap:lease:get', lease_id,
+            [statuses.CREATED, statuses.ACTIVE])
 
-        permitted.cancel()
+        lease.cancel()
 
     @staticmethod
     def _lease_get_all_authorize_filters(cdict,
@@ -207,15 +205,16 @@ class LeasesController(rest.RestController):
         }
 
         if view == 'all':
-            utils.policy_authorize('esi_leap:lease:lease_admin', cdict)
+            utils.policy_authorize('esi_leap:lease:lease_admin', cdict, cdict)
             possible_filters['owner_id'] = owner_id
             possible_filters['project_id'] = project_id
         else:
-            utils.policy_authorize('esi_leap:lease:get', cdict)
+            utils.policy_authorize('esi_leap:lease:get_all', cdict, cdict)
 
             if owner_id:
                 if cdict['project_id'] != owner_id:
-                    utils.policy_authorize('esi_leap:lease:lease_admin', cdict)
+                    utils.policy_authorize('esi_leap:lease:lease_admin', cdict,
+                                           cdict)
 
                 possible_filters['owner_id'] = owner_id
                 possible_filters['project_id'] = project_id
@@ -226,7 +225,7 @@ class LeasesController(rest.RestController):
                 else:
                     if project_id != cdict['project_id']:
                         utils.policy_authorize('esi_leap:lease:lease_admin',
-                                               cdict)
+                                               cdict, cdict)
                     possible_filters['project_id'] = project_id
 
         if (start_time and end_time is None) or \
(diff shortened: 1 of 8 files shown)


esi_leap/common/policy.py
```diff
@@ -47,13 +56,17 @@
         [{'path': '/leases', 'method': 'POST'}]),
     policy.DocumentedRuleDefault(
         'esi_leap:lease:get',
-        'rule:is_admin or rule:is_lessee or rule:is_owner',
+        'rule:is_admin or rule:is_lease_owner or rule:is_lease_lessee',
```
DanNiESh, line 59: When retrieving a lease, what's the difference between 'rule:is_lease_owner or rule:is_lease_lessee' and 'rule:is_lessee or rule:is_owner'? I think the lessee or owner can only get the lease they own.

esi_leap/api/controllers/v1/lease.py
```diff
@@ -207,15 +205,16 @@ def _lease_get_all_authorize_filters(cdict,
         }
 
         if view == 'all':
```
DanNiESh, line 207: What is the 'view' parameter?

esi_leap/common/policy.py
tzumainn: Good question! These policy rules are defined near the top of the file: is_lessee and is_owner means that the user has the lessee or owner role, while is_lease_owner/is_lease_lessee means that the user's project_id matches the owner or lessee of the lease.

Previously the code that ensured that a lessee or owner can only get the lease they own was very specifically coded with several if statements in the controller. This change pushes that check to oslo.policy.

esi_leap/api/controllers/v1/lease.py
tzumainn: Oh, that's a CLI parameter that a user using the CLI can set. By default the CLI will return only live offers and leases. The '-view all' option tells the controller to send *all* offers/leases instead (possibly for generating a report).

esi_leap/common/policy.py
DanNiESh: Gocha!

esi_leap/api/controllers/v1/lease.py
DanNiESh: Got it.

esi_leap/common/policy.py
```diff
@@ -29,6 +29,15 @@
     policy.RuleDefault('is_lessee',
                        'role:lessee or role:esi_leap_lessee',
                        description='Lessee API access'),
+    policy.RuleDefault('is_offer_owner',
+                       'project_id:%(offer.project_id)s',
+                       description='Owner of offer'),
```
samuel-zuk, line 34: Is there a reason why there's no `'is_offer_lessee`? I do know lessees can be specified when creating an offer, does this not factor into any policy decisions?

esi_leap/common/policy.py
```diff
@@ -75,17 +88,21 @@
     policy.DocumentedRuleDefault(
         'esi_leap:offer:get',
         'rule:is_admin or rule:is_owner or rule:is_lessee',
```
samuel-zuk, line 90: If the `is_offer_lessee` rule existed I could see it being useful here. (example: owner specifies a project id that does not have the lessee role as a lessee for a newly created offer)

esi_leap/common/policy.py
```diff
@@ -75,17 +88,21 @@
...
+        [{'path': '/offers/{offer_ident}', 'method': 'GET'}]),
+    policy.DocumentedRuleDefault(
+        'esi_leap:offer:get_all',
+        'rule:is_admin or rule:is_owner or rule:is_lessee',
+        'Retrieve multiple offers',
+        [{'path': '/offers', 'method': 'GET'}]),
     policy.DocumentedRuleDefault(
         'esi_leap:offer:delete',
-        'rule:is_admin or rule:is_owner',
+        'rule:is_admin or rule:is_offer_owner',
```
samuel-zuk, line 100: Was the previous behavior really that any owner could delete any offer? That sounds... kind of bad lol

esi_leap/tests/api/controllers/v1/test_utils.py
```diff
@@ -656,6 +476,49 @@ def test_check_resource_lease_admin_bad_time(self,
         self.assertIsNone(parent_lease_uuid)
 
 
+class TestOfferPolicyAndRetrieveUtils(testtools.TestCase):
+
```
samuel-zuk, line 480: Incredibly nitpicky style detail but is there a reason why there's a newline after the class/function declarations in these test cases?

samuel-zuk (commented):
Looks good!

esi_leap/common/policy.py
tzumainn: Ah, yes - it's because I'm not certain how easy it is to create a policy rule that says "if %(offer.lessee_id) is None or project_id:%(offer.project_id)s"

tzumainn: Now that I think about it - I was going to look into this anyway, but then I remembered: the offer lessee_id actually represent an entire project tree. Oslo policy has relatively simple matching rules that don't accommodate for that.

esi_leap/common/policy.py
tzumainn: Kind of! Previously, the policy check for delete would pass as long as you had the owner role; then an additional check in the controller would verify that you were the owner of that offer. This change gets rid of the need for that second check.

esi_leap/tests/api/controllers/v1/test_utils.py
tzumainn: I think it's because what a lot of OpenStack tests do, and because our linter doesn't complain :)

esi_leap/tests/api/controllers/v1/test_utils.py
```diff
@@ -410,57 +208,77 @@ def test_get_offer_name_available(self,
...
+    def test_get_lease_uuid_available(self,
+                                      mock_lease_get,
+                                      mock_is_uuid_like):
 
         mock_is_uuid_like.return_value = True
         mock_lease_get.return_value = test_lease
 
-        res = utils.get_lease_authorized(test_lease.uuid,
-                                         lessee_ctx.to_policy_values())
+        res = utils.get_lease(test_lease.uuid, [statuses.CREATED])
```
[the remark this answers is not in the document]
tzumainn: Ah, leases don't have an available status; only offers do. It's because leases are never claimed or used by others; they're simply created, fulfilled, and then expired.

esi_leap/tests/api/controllers/v1/test_utils.py
```diff
@@ -410,57 +208,77 @@ def test_get_offer_name_available(self,
...
+                                      mock_is_uuid_like):
 
         mock_is_uuid_like.return_value = False
         mock_lease_get_all.return_value = [test_lease]
 
-        res = utils.get_lease_authorized(
-            test_lease.name,
-            lessee_ctx.to_policy_values(),
-            [statuses.CREATED, statuses.ACTIVE])
+        res = utils.get_lease(test_lease.name, statuses.AVAILABLE)
```
DanNiESh, line 266: So this status should be CREATED?

tzumainn: Yep - nice catch! I'll push a fix.

DanNiESh (approved):
Looks good to me!

PR #85, opened 2021-07-26
Include owner, offer, and lease information in node list

betaredex:


esi_leap/api/controllers/v1/node.py
```diff
@@ -47,6 +56,16 @@ def get_all(self):
         nodes = ironic.get_node_list(context)
 
         node_collection = NodeCollection()
-        node_collection.nodes = [Node(name=n.name) for n in nodes]
+
+        for node in nodes:
+            offers = offer_obj.Offer.get_all({"resource_uuid": node.uuid},
+                                             context)
```
tzumainn, line 62: This will get all offers, including any expired ones - and since you only show the first one (which is reasonable), you'll probably end up showing an expired one. You may want to add more filters here to filter by status.

esi_leap/api/controllers/v1/node.py
```diff
@@ -47,6 +56,16 @@ def get_all(self):
...
+            n = Node(name=node.name, uuid=node.uuid, owner=node.owner)
+            if offers:
+                n.offer_uuid = offers[0].uuid
```
tzumainn, line 65: Showing one is reasonable, but I think it should either be:

* show a current offer, in which case you need to further update the offer filter to limit offers to a current time
* show all active and future offers

Either way, I think adding the start/end time would be useful.

esi_leap/api/controllers/v1/node.py
```diff
@@ -47,6 +56,16 @@ def get_all(self):
...
+            if "lease_uuid" in node.properties:
+                n.lease_uuid = node.properties["lease_uuid"]
```
tzumainn, line 67: This only shows the current lease; if you decide to only show the current offer, then this is enough; if you decide to show active and future offers, then you should do the same with the lease. Either way, adding in the lease's start/end time would be useful as well!

tzumainn (commented):
Oh, I forgot - instead of displaying the project id for the owner, you can use the function in https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/common/keystone.py#L64 to display the project name of the owner. Since there are potentially multiple nodes, I suggest the version where you pass in a project_list, as in https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/api/controllers/v1/lease.py#L119

esi_leap/api/controllers/v1/node.py
betaredex: Added a status filter

esi_leap/api/controllers/v1/node.py
betaredex: I decided to show one for now, but I added filters to make sure its only the currently available one. I can add a future offer uuids field if needed.

esi_leap/api/controllers/v1/node.py
```diff
@@ -19,24 +20,34 @@
 from esi_leap.api.controllers import base
 from esi_leap.api.controllers import types
 from esi_leap.common import ironic
+from esi_leap.common import keystone
 import esi_leap.conf
 
+from esi_leap.objects import offer as offer_obj
```
tzumainn, line 26: Can you get rid of this newline?

tzumainn (changes requested):
Generally looks good! For this iteration, I'd suggest one more change: get rid of the node UUID (we're encouraging people to use names), and add in the lessee project.

tzumainn (approved):
Looks good!

PR #87, opened 2021-07-28
Removed owner change code

tzumainn:
The owner change requirement has been re-framed to no longer require
actual owner changes.


esi_leap/resource_objects/ironic_node.py
```diff
@@ -92,14 +92,6 @@ def expire_lease(self, lease):
         if state == "active":
             get_ironic_client().node.set_provision_state(self._uuid, "deleted")
 
-    def set_owner(self, owner_id):
```
ljmcgann, line 95: its probably fine but what did this piece of code do?

ljmcgann (approved):
lgtm

esi_leap/resource_objects/ironic_node.py
tzumainn: It changed the owner of the node - which is pretty dangerous for several reasons.

PR #88, opened 2021-07-29
Fix error when controller attempts to retrieve project name

tzumainn:
The controller will attempt to retrieve a project name even if
the project id is not set - for example, if an offer does not have
a lessee specified. This patch fixes the issue.

### DIFF
--- a/esi_leap/common/keystone.py
+++ b/esi_leap/common/keystone.py
@@ -63,12 +63,16 @@ def get_project_list():
 
 def get_project_name(project_id, project_list=None):
     project_name = ''
-    if project_list is None:
-        project = get_keystone_client().projects.get(project_id)
-    else:
-        project = next((p for p in project_list
-                        if getattr(p, 'id') == project_id),
-                       None)
-    if project:
-        project_name = project.name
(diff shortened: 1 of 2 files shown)


ljmcgann:
What would a user have to do to make this issue pop up? Im slightly confused

tzumainn:
> What would a user have to do to make this issue pop up? Im slightly confused

An owner can now create an offer and specifies a lessee: this limits the offer to that project tree. If they do so, they won't run into this problem; however, if they don't specify a lessee, then as of now the code will try and find a name for that empty project.

ljmcgann:
Not specifying a lessee when creating an offer is the original workflow for offer creation right? Or is this slightly different? Has that been broken this entire time then?

tzumainn:
> Not specifying a lessee when creating an offer is the original workflow for offer creation right? Or is this slightly different? Has that been broken this entire time then?

Yep, that's correct; specifying a lessee is relatively new. And the bit that broke it is something that came after: returning the project name as well as the project UUID.

ljmcgann (approved):
lgtm

PR #89, opened 2021-08-04
Update the lease returned when claiming an offer to include names

tzumainn:
In order to allow the lease to include the names, we have to move
lease/offer controller functions into utils.

### DIFF
--- a/esi_leap/api/controllers/v1/utils.py
+++ b/esi_leap/api/controllers/v1/utils.py
@@ -139,3 +139,31 @@ def check_offer_lessee(cdict, offer):
         resource_policy_authorize(
             'esi_leap:offer:offer_admin',
             cdict, cdict, 'offer', offer.uuid)
+
+
+def offer_get_dict_with_added_info(offer,
+                                   project_list=None,
+                                   node_list=None):
+    availabilities = offer.get_availabilities()
+    resource = offer.resource_object()
+
+    o = offer.to_dict()
+    o['availabilities'] = availabilities
+    o['project'] = keystone.get_project_name(offer.project_id,
+                                             project_list)
+    o['lessee'] = keystone.get_project_name(offer.lessee_id,
+                                            project_list)
+    o['resource'] = resource.get_resource_name(node_list)
+    return o
+
+
+def lease_get_dict_with_added_info(lease, project_list=None, node_list=None):
+    resource = lease.resource_object()
+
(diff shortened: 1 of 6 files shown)


betaredex:
Looks good to me!

esi_leap/api/controllers/v1/utils.py
```diff
@@ -139,3 +139,31 @@ def check_offer_lessee(cdict, offer):
         resource_policy_authorize(
             'esi_leap:offer:offer_admin',
             cdict, cdict, 'offer', offer.uuid)
+
+
+def offer_get_dict_with_added_info(offer,
```
samuel-zuk, line 144: Does this have to be on multiple lines? I only ask because the function declaration for `lease_get_dict_with_added_info()` is on one line

esi_leap/api/controllers/v1/utils.py
```diff
@@ -139,3 +139,31 @@ def check_offer_lessee(cdict, offer):
...
+    o['lessee'] = keystone.get_project_name(offer.lessee_id,
+                                            project_list)
+    o['resource'] = resource.get_resource_name(node_list)
+    return o
+
+
+def lease_get_dict_with_added_info(lease, project_list=None, node_list=None):
+    resource = lease.resource_object()
+
+    lease_dict = lease.to_dict()
```
samuel-zuk, line 163: Another incredibly picky semantic thing, since this variable is named `lease_dict`, should the corresponding var in `offer_get_dict...` be named `offer_dict`? (or this variable be named `l`?)

samuel-zuk (commented):
Looks good!

DanNiESh (approved):
Looks good to me!

esi_leap/api/controllers/v1/utils.py
tzumainn: It wouldn't hurt, you're right! But I'd consider such reconciliation to be out of the scope of this PR. Generally speaking we want to make sure each PR does one thing while avoiding scope creep. Otherwise the PR either ends up with a commit message that covers many separate ideas, or - more likely - the commit message actually misses some of the contents of the patch.

esi_leap/api/controllers/v1/utils.py
tzumainn: Probably not, but as long as the linter doesn't complain it should be fine!

PR #90, opened 2021-08-04
Add future offers and future leases to the node list display

betaredex:


esi_leap/api/controllers/v1/node.py
```diff
@@ -21,8 +21,10 @@
 from esi_leap.api.controllers import types
 from esi_leap.common import ironic
 from esi_leap.common import keystone
+from esi_leap.common import statuses
 import esi_leap.conf
 from esi_leap.objects import offer as offer_obj
+from esi_leap.objects import lease as lease_obj
```
tzumainn, line 27: I'm not sure why the CI doesn't pick this up, but can you move the lease import above the offer import? 'tox -epep8' complains about imports not being in alphabetical order.

esi_leap/api/controllers/v1/node.py
```diff
@@ -64,14 +68,27 @@ def get_all(self):
...
+            f_lease_uuids = ' '.join([o.uuid for o in f_leases])
+
             n = Node(name=node.name, uuid=node.uuid,
-                     owner=keystone.get_project_name(node.owner, project_list))
-            if offers:
-                n.offer_uuid = offers[0].uuid
+                     owner=keystone.get_project_name(node.owner, project_list),
+                     future_offers=future_offers,
+                     future_leases=f_lease_uuids)
+            if 'current_offer' in locals():
```
tzumainn, line 90: I think the more standard way to do this is to define ``current_offer = None`` up above, and then check ``if current_offer:`` here.

betaredex: Done

esi_leap/api/controllers/v1/node.py
betaredex: Done

tzumainn (approved):
Looks reasonable! I think the final display should probably end up showing times and the like, but that can be iterated upon based on feedback from the demo.

PR #92, opened 2022-04-25
Added error status for offers and leases

tzumainn:
Adds error status for offers and leases, and exception handling
around manager services.


DanNiESh (approved):
Looks good to me!

PR #93, opened 2022-05-25
Update CI to use GitHub Actions

tzumainn:


mnbatra (approved):
Looks good

ljmcgann (commented):
Why are we able to remove all these files? I get that the functionality is being replaced but I don't know enough about the tooling here to say if this does what we want

tzumainn:
The files being removed are specifically for Travis CI; we're moving from that to GitHub actions, so those travis files should be removed as they'll no longer do anything.

ljmcgann:
lgtm

ljmcgann (approved):

PR #101, opened 2022-07-06
Add alembic db migration

DanNiESh:
Add esi-leap-dbsync <version| revision|upgrade|downgrade> commands. Tested manually and all of them work.


tzumainn (approved):
Looks good - thanks!

PR #102, opened 2022-07-08
Added 'wait' state for leases

tzumainn:
A lease can fail to be fulfilled or expired simply because the node
is not in the correct state. In such cases, we change the state of the
lease to 'wait' and simply try again later.

### DIFF
--- a/esi_leap/manager/service.py
+++ b/esi_leap/manager/service.py
@@ -50,6 +50,8 @@ class ManagerService(service.Service):
         self.tg.add_timer(EVENT_INTERVAL, self._fulfill_leases)
         LOG.info("Starting _expire_leases periodic job")
         self.tg.add_timer(EVENT_INTERVAL, self._expire_leases)
+        LOG.info("Starting _cancel_leases periodic job")
+        self.tg.add_timer(EVENT_INTERVAL, self._cancel_leases)
         LOG.info("Starting _expire_offers periodic job")
         self.tg.add_timer(EVENT_INTERVAL, self._expire_offers)
 
@@ -61,7 +63,8 @@ class ManagerService(service.Service):
     def _fulfill_leases(self):
         LOG.info("Checking for leases to fulfill")
         leases = lease_obj.Lease.get_all(
-            {'status': [statuses.CREATED]}, self._context)
+            {'status': [statuses.CREATED, statuses.WAIT_FULFILL]},
+            self._context)
         now = timeutils.utcnow()
         for lease in leases:
             if lease.start_time <= now and now <= lease.end_time:
@@ -69,14 +72,16 @@ class ManagerService(service.Service):
                     LOG.info("Fulfilling lease %s", lease.uuid)
                     lease.fulfill(self._context)
                 except Exception as e:
-                    LOG.info("Error fulfilling lease: %s", e)
+                    LOG.info("Error fulfilling lease; setting to ERROR: %s", e)
                     lease.status = statuses.ERROR
                     lease.save()
 
     def _expire_leases(self):
         LOG.info("Checking for expiring leases")
         leases = lease_obj.Lease.get_all(
-            {'status': [statuses.ACTIVE, statuses.CREATED]}, self._context)
+            {'status': [statuses.ACTIVE, statuses.CREATED,
+                        statuses.WAIT_EXPIRE, statuses.WAIT_FULFILL]},
+            self._context)
         now = timeutils.utcnow()
         for lease in leases:
             if lease.end_time <= now:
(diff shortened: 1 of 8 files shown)


DanNiESh:
Hi! Do we still need the 'error' status? I think it's replaced by 'wait' status and never been used.

tzumainn:
> Hi! Do we still need the 'error' status? I think it's replaced by 'wait' status and never been used.

Good point! I think it's still used if an offer somehow fails to be expired.

samuel-zuk:
> > Hi! Do we still need the 'error' status? I think it's replaced by 'wait' status and never been used.
> 
> Good point! I think it's still used if an offer somehow fails to be expired.

yep, esi_leap/manager/service.py:107 expires offers with statuses.ERROR

esi_leap/manager/service.py
```diff
@@ -70,13 +70,14 @@ def _fulfill_leases(self):
                     lease.fulfill(self._context)
                 except Exception as e:
                     LOG.info("Error fulfilling lease: %s", e)
-                    lease.status = statuses.ERROR
+                    lease.status = statuses.WAIT
```
samuel-zuk, line 73: If the reason the lease failed to be fulfilled/expired is due to the node being in the wrong state, would we want to distinguish that failure mode (by setting status = WAIT) from other possible failure modes (in which we should set status = ERROR)?

DanNiESh:
> > > Hi! Do we still need the 'error' status? I think it's replaced by 'wait' status and never been used.
> > 
> > 
> > Good point! I think it's still used if an offer somehow fails to be expired.
> 
> yep, esi_leap/manager/service.py:107 expires offers with statuses.ERROR

I see!

DanNiESh (approved):
Thanks! LGTM

esi_leap/manager/service.py
```diff
@@ -70,13 +70,14 @@ def _fulfill_leases(self):
                     lease.fulfill(self._context)
                 except Exception as e:
                     LOG.info("Error fulfilling lease: %s", e)
```
samuel-zuk, line 72: Also probably out of scope for this PR but would we want to include the exception type in the error log? If we have an exception with a name like OutOfMemoryException with a faultstring that's blank or unhelpful without the class name, we'd be leaving out key context.

tzumainn:
Now that I think about it, I should probably put the "WAIT" assignment higher up, in the lease's fulfill/expire method directly, since leases can be expired by an expiring offer as well. Let me make that change; I'll let you guys know when this is ready to be reviewed again. Thanks for the comments so far!

esi_leap/tests/api/controllers/v1/test_lease.py
```diff
@@ -729,7 +729,7 @@ def test_lease_get_all_status(self):
 
         expected_filters = {
             'project_or_owner_id': 'adminid',
-            'status': ['created', 'active', 'error']
+            'status': ['created', 'active', 'error', 'wait']
```
samuel-zuk, line 732: Not really your fault here since this problem was pre-existing, but these should probably be references to statuses.CREATED/ACTIVE/etc. rather than their string values.

samuel-zuk (commented):
Looks good! Just got some semantic qualms.

esi_leap/manager/service.py
tzumainn: Fair question! The answer right now is I think I'd prefer to be in WAIT as much as possible (and thus eligible for a retry), as any error in the lease handling has to be remediated somehow.

tzumainn:
Okay, I've updated the PR so that a lease fulfill/cancel/expire can trigger separate wait states, which are then picked up by periodic jobs. Let me know what you think!

esi_leap/tests/api/controllers/v1/test_lease.py
tzumainn: Yeah, you're completely right - fixed!

esi_leap/manager/service.py
tzumainn: Ah, you're right - I'm going to open up a new issue for this one for someone else to fix!

DanNiESh:
Do we want to add 'WAIT_EXPIRE', 'WAIT_FULFILL' in lease_obj.Lease.get_all here? https://github.com/CCI-MOC/esi-leap/blob/769d8b143c1b74961f1b0e8a14703816b4a0709f/esi_leap/objects/offer.py#L150

tzumainn:
> Do we want to add 'WAIT_EXPIRE', 'WAIT_FULFILL' in lease_obj.Lease.get_all here?
> 
> https://github.com/CCI-MOC/esi-leap/blob/769d8b143c1b74961f1b0e8a14703816b4a0709f/esi_leap/objects/offer.py#L150

You are completely correct! I've added a new array of statuses, which is used in various places. Good catch!

DanNiESh (approved):
Looks good to me!

PR #105, opened 2022-07-12
Made changes to remove restriction for delete offer/lease as per the feedback on previous closed PR

rd4398:


esi_leap/manager/service.py
```diff
@@ -80,7 +80,8 @@ def _expire_leases(self):
         LOG.info("Checking for expiring leases")
         leases = lease_obj.Lease.get_all(
             {'status': [statuses.ACTIVE, statuses.CREATED,
-                        statuses.WAIT_EXPIRE, statuses.WAIT_FULFILL]},
+                        statuses.WAIT_EXPIRE, statuses.WAIT_FULFILL,
+                        statuses.OFFER_CAN_DELETE]},
```
tzumainn, line 84: Ah, we don't want to add OFFER_CAN_DELETE to an array of lease statuses.

esi_leap/common/statuses.py
```diff
@@ -20,4 +20,7 @@
 WAIT_EXPIRE = 'wait expire'
 WAIT_FULFILL = 'wait fulfill'
 
+OFFER_CAN_DELETE = ('available', 'error')
```
tzumainn, line 23: You'll want to use AVAILABLE and ERROR here instead of duplicating strings. Any reason we're using '(' instead of '['?

esi_leap/common/statuses.py
```diff
@@ -20,4 +20,7 @@
...
+LEASE_CAN_DELETE = ['active', 'created', 'error']
```
tzumainn, line 24: LEASE_CAN_DELETE and CANCELLABLE_LEASE_STATUSES can probably be combined (I like your variable name better!). Just make sure it contains all statuses from both.

tzumainn (commented):
A few minor changes needed, but I think it's getting close!

esi_leap/common/statuses.py
rd4398: Yeah, sam_z suggested that since there are only two values, it should be a tuple rather than list so I changed '[' to '('.
Should I changed it back?

esi_leap/common/statuses.py
rd4398: Okay, will do!

esi_leap/common/statuses.py
tzumainn: I don't think it's a big deal either way; you can just leave it!

PR #106, opened 2022-07-13
Grab lessee from node attributes and code improvements

rd4398:
As per our discussions yesterday, I have made changes to the code... I am not sure whether my changes improve the code significantly and whether these changes will affect other pieces or not. For now I have focused on reducing number of API calls...

Please let me know if any changes are required!

### DIFF
--- a/esi_leap/api/controllers/v1/lease.py
+++ b/esi_leap/api/controllers/v1/lease.py
@@ -180,7 +180,7 @@ class LeasesController(rest.RestController):
 
         lease = utils.check_lease_policy_and_retrieve(
             request, 'esi_leap:lease:get', lease_id,
-            [statuses.CREATED, statuses.ACTIVE])
+            statuses.LEASE_CAN_DELETE)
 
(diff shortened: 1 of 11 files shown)


tzumainn:
Hi! It looks like this PR has two different commits intended for two different PRs; when you create a new PR, make sure you create a new branch so that each branch is isolated to a single PR.

rd4398:
Ah! I did create a branch "rohand-issue#244" for the second commit...looks like the "master" branch is not considered a different one...
What can be done now?

tzumainn:
I'm guessing you created a branch off of your previous branch, and not the master branch? In general, the master branch of your fork should always correspond to the master of the main repository to avoid issues like this.

If that's already the case, then you can just 'git checkout master', create a new branch, and then move your commit there; one way would be to do a 'git diff HEAD~1 > lessee.patch' from this branch, and then apply that patch to the new branch.

rd4398:
Okay, will take a look and fix it..

PR #107, opened 2022-07-13
Changes done as per the feedback related to issue #263

rd4398:
I created this new PR which has changes related to removing restrictions for delete lease/offer along with the recent feedback that you gave to me!

Please let me know if any other changes are required!


esi_leap/common/statuses.py
```diff
@@ -20,4 +20,7 @@
 WAIT_EXPIRE = 'wait expire'
 WAIT_FULFILL = 'wait fulfill'
 
-CANCELLABLE_LEASE_STATUSES = [ACTIVE, CREATED, WAIT_FULFILL]
+OFFER_CAN_DELETE = (AVAILABLE, ERROR)
+LEASE_CAN_DELETE = [ACTIVE, CREATED, ERROR, WAIT_FULFILL]
+
+#    CANCELLABLE_LEASE_STATUSES = [ACTIVE, CREATED, WAIT_FULFILL]
```
tzumainn, line 26: No need to comment this out; you can just remove it!

tzumainn (changes requested):
One last minor issue, then I think it's probably good!

esi_leap/common/statuses.py
rd4398: okay!

rd4398:
done! 
please check!

tzumainn:
It looks good! One more thing: you'll want to change your commit message. That's what's going into the historical record, so it should be very specific as to what the commit actually does. For example:

```
Allow offers/leases in ?? states to be deleted

This change allows offers and leases in ?? states to be deleted instead of hanging around
forever.
```

rd4398:
Ah okay! I will change it

tzumainn (approved):
Looks good - thanks!

PR #113, opened 2022-07-21
Cleanup and optimization within esi_leap codebase

samuel-zuk:
Some of the code in the codebase contains redundancies, has missing edge
case coverage, or is implemented in ways that could be more efficient.
This commit addresses this, implementing the following changes:

- esi_leap/api/controllers/base.py:to_dict:20-28:
  - Remove redundant getattr() calls, as the previous implementation
    called getattr() twice for every field in self.fields and contained
    a redundant hasattr() call.
  - Benchmarked as potentially twice as fast.

- esi_leap/api/controllers/types.py:41:
  - Removed the class variable 'next', as it is seemingly unused and
    since defining 'next' may cause problems with iterators.

- esi_leap/api/controllers/types.py:collection:55-57:
  - Previous implementation instantiated a list object that would be
    immediately discarded.
  - New implementation appears to be ~10% faster.

- esi_leap/api/controllers/v1/lease.py:
      _lease_get_all_authorize_filters:193-195:
  - Old implementation preformed 3 conditional checks if status is not
    None or 'any', this performs 1 check in that case and 2 checks max.

- esi_leap/api/controllers/v1/lease.py:
      _lease_get_all_authorize_filters:200:246:
- esi_leap/api/controllers/v1/offer.py:get_all:148-163:
  - The 'possible_filters' dict is redundant-- we don't need to create 2
    objects. This renames 'possible_filters' to 'filters' and deletes
    dict entries with null values at the end instead.

- esi_leap/api/controllers/v1/lease.py:
      _lease_get_all_authorize_filters:233-241:
- esi_leap/api/controllers/v1/offer.py:get_all:117-135:
  - Previous validation of start/end datetime objects performed 5-8
    checks/comparisons, this reduces that to 4.
  - Also cleans up formatting a bit in the process.

- esi_leap/api/controllers/v1/utils.py:get_offer:63-72:
  - Changed variable name to be consistent with get_lease() function.

- esi_leap/api/controllers/v1/utils.py:get_offer:66-70:
- esi_leap/api/controllers/v1/utils.py:get_lease:86-90:
  - Previous implementation always performed 2 checks, this performs 1-2
    checks with the second check only taking place if an exception is to
    be thrown

- esi_leap/common/ironic.py:get_node_name:51-60:
- esi_leap/common/keystone.py:get_project_name:66-76:
  - Eliminates redundant resource_name variable.

- esi_leap/common/keystone.py:get_parent_project_id_tree:41-45:
  - Previous implementation performed multiple calls to
    get_keystone_client(), this performs only one.

- esi_leap/db/api.py:decorator:63-69:
  - Previous version will raise error if the output of res is a non-list
    iterable (e.g. a tuple)

- esi_leap/objects/base.py:to_dict:43:
  - _attr_as_dict() defined but doesn't seem to be used. Removing it
    doesn't seem to break anything.

- esi_leap/objects/lease.py:create:68-108:
- esi_leap/objects/lease.py:cancel:123-141:
- esi_leap/objects/lease.py:fulfill:153-168:
- esi_leap/objects/lease.py:expire:183-201:
- esi_leap/objects/offer.py:create:122-152:
- esi_leap/objects/offer.py:cancel:162-166:
- esi_leap/objects/offer.py:expire:176-180:
  - Previous implementation, which created a locally-scoped decorated
    function which was immediately called is slightly slower, needs
    slightly more overhead, and is clunky. Replaced with a context
    manager.
- esi_leap/common/utils.py:16:
  - Refactoring described above removed all instances of the
    'synchronized' decorator, so it can be removed here too.
- esi_leap/objects/offer.py:cancel:162-163:
- esi_leap/objects/offer.py:expire:176-177:
  - Added external=True to lock creation command.

- esi_leap/resource_objects/ironic_node.py:expire_lease:81-84:
  - Eliminated redundant lease.get_lease_uuid() call.
- esi_leap/tests/resource_objects/test_ironic_node.py:test_expire_lease:
  - Updated to reflect elimination of redundant get_lease_uuid call.

- esi_leap/tests/api/base.py:setUp:39:
  - Patcher is now stopped upon test case teardown.

### DIFF
--- a/esi_leap/api/controllers/base.py
+++ b/esi_leap/api/controllers/base.py
@@ -17,7 +17,10 @@ from wsme import types as wtypes
 class ESILEAPBase(wtypes.Base):
 
     def to_dict(self):
-        return dict(
-            (k, getattr(self, k))
-            for k in self.fields
-            if hasattr(self, k) and getattr(self, k) != wsme.Unset)
+        esi_leap_dict = {}
+        if self.fields:
+            for key in self.fields:
+                val = getattr(self, key, wsme.Unset)
+                if val != wsme.Unset:
+                    esi_leap_dict[key] = val
+        return esi_leap_dict
--- a/esi_leap/api/controllers/types.py
+++ b/esi_leap/api/controllers/types.py
@@ -39,8 +39,6 @@ class JsonType(wtypes.UserType):
 
 class Collection(wtypes.Base):
 
-    next = wtypes.text
-
     @property
     def collection(self):
         return getattr(self, self._type)
@@ -54,7 +52,7 @@ class Collection(wtypes.Base):
             return wtypes.Unset
 
         url = url or self._type
-        q_args = ''.join(['%s=%s&' % (key, kwargs[key]) for key in kwargs])
+        q_args = ''.join(['%s=%s&' % item for item in kwargs.items()])
         next_args = '?%(args)slimit=%(limit)d&marker=%(marker)s' % {
             'args': q_args, 'limit': limit,
             'marker': getattr(self.collection[-1], 'uuid')}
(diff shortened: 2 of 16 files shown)


tzumainn (approved):
Looks good to me!

PR #114, opened 2022-07-21
Ensured exception handling logs exception type

samuel-zuk:
If an operation on an object fails with an exception, the exception
message will be displayed, but currently, the type is excluded. This is
not ideal-- if an exception like OutOfMemory('64G/64G') was raised, the
message may not provide enough context to solve the root problem. This
patch addresses this.

Addresses #267.


tzumainn (approved):
Looks good - thanks!

PR #115, opened 2022-07-21
Removed ResourceObjectFactory

samuel-zuk:
The ResourceObjectFactory class does not provide a meaningful layer of
abstraction beyond the static method function it encapsultes. Its
implementation simply calls another object constructor, and in the case
of Ironic nodes, handles data validation that is better suited to the
IronicNode constructor. This redundancy adds needless complexity to the
ESI-Leap software and was factored out in this patch.

Changes made:
- Removed ResourceObjectFactory
- Moved the get_resource_object helper function and the RESOURCE_TYPES
  array to the module level.
- Added get_type helper function to get the object type from a resource
  name string
- `esi_leap.resource_objects` now builds the RESOURCE_TYPES array
  programatically via checking the subclasses of the base class.
- Instances of the ResourceObjectFactory were changed to use the
  get_resource_object helper function instead and tests were updated to
  match.
- Added new tests for the new helper functions.


tzumainn (approved):
Makes sense to me - thanks!

PR #117, opened 2022-07-26
Fix for offer status filters

tzumainn:
The offer status filter was changed to be an array; this updates
parameters to reflect that.


DanNiESh (commented):
One question: is it necessary to update status to [status] in unit tests?

tzumainn:
> One question: is it necessary to update status to [status] in unit tests?

It's actually not, since the tests apparently only tested the array/None cases! That's a bit of a gap so I added a test to test a single status. Good catch!

DanNiESh (approved):
Looks good! Thanks!

PR #118, opened 2022-07-26
Fix offer claim status field

tzumainn:


DanNiESh:
I found here: https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/api/controllers/v1/lease.py#L182, https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/api/controllers/v1/offer.py#L239. These might be similar issues?

tzumainn:
> I found here: https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/api/controllers/v1/lease.py#L182, https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/api/controllers/v1/offer.py#L239. These might be similar issues?

Ah, nope! Those are both arrays already: https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/common/statuses.py

DanNiESh (approved):
LGTM!

PR #120, opened 2022-08-19
Add state fields to node controller

tzumainn:
When displaying node information, it's often useful to be able to
see the node's provisioning state and maintenance info.


DanNiESh (approved):
Thanks! Looks good!

PR #122, opened 2022-08-25
Ensure an IronicNode does not retrieve a node from Ironic more than once

tzumainn:


esi_leap/resource_objects/ironic_node.py
```diff
@@ -34,34 +34,33 @@ class IronicNode(base.ResourceObjectInterface):
...
-        node = get_ironic_client().node.get(self._uuid)
-        return node.properties.get('lease_uuid', None)
+        return self._get_node().properties.get('lease_uuid', None)
 
     def get_project_id(self):
-        node = get_ironic_client().node.get(self._uuid)
-        return node.lessee
+        return self._get_node().lessee
 
     def get_node_config(self):
```
DanNiESh, line 55: Not relevant to this PR, but I don't see this function is called anywhere.

esi_leap/resource_objects/ironic_node.py
```diff
@@ -94,10 +93,14 @@ def expire_lease(self, lease):
...
+        state = self._get_node().provision_state
         if state == 'active':
             get_ironic_client().node.set_provision_state(self._uuid, 'deleted')
 
     def resource_admin_project_id(self):
-        node = get_ironic_client().node.get(self._uuid)
-        return node.owner
+        return self._get_node().owner
+
+    def _get_node(self, resource_list=None):
```
DanNiESh, line 103: How do we prevent that the cached node we are using is out-dated? I guess there might be a very small possibility that it happens.

DanNiESh: Do we need a test for  _get_node()?

esi_leap/manager/service.py
```diff
@@ -104,7 +104,7 @@ def _cancel_leases(self):
         for lease in leases:
             try:
                 LOG.info('Cancelling lease %s', lease.uuid)
-                lease.cancel(self._context)
```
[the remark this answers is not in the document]
tzumainn: Haha, I'm actually not sure; it's something to cleanup, but in the meantime this statement was causing errors.

esi_leap/resource_objects/ironic_node.py
tzumainn: Weird, you're right - I'm not sure what it was used for either!

esi_leap/resource_objects/ironic_node.py
tzumainn: Well, this should only get cached within a single request, so that's fine. I should definitely have added tests - they should be included now!

tzumainn:
@DanNiESh I've addressed your comments - let me know what you think!

DanNiESh (approved):
Looks good to me!

PR #123, opened 2022-09-08
Parallelize API requests for esi-leap

rd4398:
I explored threads approach and gave a shot with concurrent.futures module of python which uses ThreadPoolExecutor. This is similar to traditional threading module but easy to implement and requires less lines of code to achieve what we want. Here are the links I referred to:

Youtube (Start from 15:00 to 20:00) : https://www.youtube.com/watch?v=IEEhzQoKtQU
Python Docs: https://docs.python.org/3/library/concurrent.futures.html

Please let me know if this approach is what I can use and whether it fulfills our requirement.

PS: I am getting the following error even though I have defined the function:

`File "/home/rohand/esi-leap/.tox/py3/lib/python3.10/site-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)

      File "/home/rohand/esi-leap/esi_leap/tests/api/controllers/v1/test_node.py", line 52, in test_get_all
    data = self.get_json('/nodes')

      File "/home/rohand/esi-leap/esi_leap/tests/api/base.py", line 73, in get_json
    response = self.app.get(full_path,

      File "/home/rohand/esi-leap/.tox/py3/lib/python3.10/site-packages/webtest/app.py", line 324, in get
    return self.do_request(req, status=status,

      File "/home/rohand/esi-leap/.tox/py3/lib/python3.10/site-packages/webtest/app.py", line 646, in do_request
    self._check_status(status, res)

      File "/home/rohand/esi-leap/.tox/py3/lib/python3.10/site-packages/webtest/app.py", line 675, in _check_status
    raise AppError(

    webtest.app.AppError: Bad response: 500 Internal Server Error (not 200 OK or 3xx redirect for http://localhost/v1/nodes)
b'{"faultcode": "Server", "faultstring": "name \'get_all_nodes\' is not defined", "debuginfo": null}'`

Any idea?

### DIFF
--- a/esi_leap/api/controllers/v1/lease.py
+++ b/esi_leap/api/controllers/v1/lease.py
@@ -10,6 +10,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+import concurrent.futures
 import datetime
 import http.client as http_client
 from oslo_utils import uuidutils
@@ -115,9 +116,17 @@ class LeasesController(rest.RestController):
         leases = lease_obj.Lease.get_all(filters, request)
 
         lease_collection.leases = []
+
         if len(leases) > 0:
-            project_list = keystone.get_project_list()
-            node_list = ironic.get_node_list()
+            project_list = None
+            node_list = None
+
+            with concurrent.futures.ThreadPoolExecutor() as executor:
+                f1 = executor.submit(ironic.get_node_list)
+                f2 = executor.submit(keystone.get_project_list)
+                node_list = f1.result()
+                project_list = f2.result()
+
             leases_with_added_info = [
                 Lease(**utils.lease_get_dict_with_added_info(l, project_list,
                                                              node_list))
--- a/esi_leap/api/controllers/v1/node.py
+++ b/esi_leap/api/controllers/v1/node.py
@@ -10,6 +10,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+import concurrent.futures
 from datetime import datetime
 import pecan
 from pecan import rest
@@ -63,11 +64,18 @@ class NodesController(rest.RestController):
     @wsme_pecan.wsexpose(NodeCollection)
     def get_all(self):
         context = pecan.request.context
-        nodes = ironic.get_node_list(context)
+
+        nodes = None
+        project_list = None
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            f1 = executor.submit(ironic.get_node_list, context)
+            f2 = executor.submit(keystone.get_project_list)
+            nodes = f1.result()
+            project_list = f2.result()
 
         node_collection = NodeCollection()
 
-        project_list = keystone.get_project_list()
         now = datetime.now()
 
         offers = offer_obj.Offer.get_all({'status': [statuses.AVAILABLE]},
--- a/esi_leap/api/controllers/v1/offer.py
+++ b/esi_leap/api/controllers/v1/offer.py
@@ -10,6 +10,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+import concurrent.futures
 import datetime
 import http.client as http_client
 from oslo_utils import uuidutils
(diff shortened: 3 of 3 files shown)


esi_leap/api/controllers/v1/node.py
```diff
@@ -60,14 +62,38 @@ def __init__(self, **kwargs):
...
     def get_all(self):
         context = pecan.request.context
-        nodes = ironic.get_node_list(context)
+        # t1 = threading.Thread(target=get_all_nodes)
+
+        nodes = None
+        project_list = None
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            f1 = executor.submit(get_all_nodes)
```
tzumainn, line 82: I think you can avoid having a new function and just have

```
f1 = executor.submit(ironic.get_node_list, context)
```

tzumainn (changes requested):
This approach definitely looks reasonable! Have you tested it?

esi_leap/api/controllers/v1/node.py
rd4398: Okay, will try!
Also, I am running unit tests but not sure how I can test time

esi_leap/api/controllers/v1/node.py
```diff
@@ -63,19 +64,26 @@ class NodesController(rest.RestController):
...
+
+        nodes = None
+        project_list = None
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            f1 = executor.submit(ironic.get_node_list, context)
+            f2 = executor.submit(keystone.get_project_list)
+            f3 = executor.submit(offer_obj.Offer.get_all,
+                                 {'status': [statuses.AVAILABLE]}, context)
+            f4 = executor.submit(lease_obj.Lease.get_all,
```
tzumainn, line 76: I don't think we need to do this for offers and leases, as those are direct database calls, which should be very fast. Ironic/Keystone calls are API calls, so they'll be a lot slower!

rd4398: Ah, okay!

rd4398:
I have done changes to the code and added parallel calls for keystone and ironic...
Please let me know if any more changes are required..

esi_leap/api/controllers/v1/lease.py
```diff
@@ -115,9 +116,16 @@ def get_all(self, project_id=None, start_time=None, end_time=None,
         leases = lease_obj.Lease.get_all(filters, request)
 
         lease_collection.leases = []
+        project_list = None
+        node_list = None
```
tzumainn, line 120: Minor nit: this could go inside the if statement

esi_leap/api/controllers/v1/node.py
```diff
@@ -63,19 +64,26 @@ class NodesController(rest.RestController):
...
 
         offers = offer_obj.Offer.get_all({'status': [statuses.AVAILABLE]},
                                          context)
 
         leases = lease_obj.Lease.get_all({'status': [statuses.CREATED]},
                                          context)
 
+        node_collection = NodeCollection()
+
+        now = datetime.now()
```
tzumainn, line 85: Is this a leftover debugging statement?

esi_leap/api/controllers/v1/utils.py
```diff
@@ -152,8 +153,14 @@ def offer_get_dict_with_added_info(offer, project_list=None, node_list=None):
 
     o = offer.to_dict()
     o['availabilities'] = availabilities
-    o['project'] = keystone.get_project_name(offer.project_id, project_list)
-    o['lessee'] = keystone.get_project_name(offer.lessee_id, project_list)
+    with concurrent.futures.ThreadPoolExecutor() as executor:
```
tzumainn, line 156: We actually don't need this to be parallelized; passing in project_list to keystone.get_project_name prevents multiple API calls.

esi_leap/api/controllers/v1/utils.py
```diff
@@ -163,10 +170,14 @@ def lease_get_dict_with_added_info(lease, project_list=None, node_list=None):
     resource = lease.resource_object()
 
     lease_dict = lease.to_dict()
-    lease_dict['project'] = keystone.get_project_name(lease.project_id,
-                                                      project_list)
-    lease_dict['owner'] = keystone.get_project_name(lease.owner_id,
-                                                    project_list)
+    with concurrent.futures.ThreadPoolExecutor() as executor:
```
tzumainn, line 173: Same here.

tzumainn (changes requested):
Looks good in general! Some minor issues.

esi_leap/api/controllers/v1/node.py
rd4398: It was there previously, just the position was different

rd4398:
Okay, will make changes!

rd4398:
I have done the changes...let me know if any more changes are required....

tzumainn (approved):

tzumainn:
Looks good - thanks!

PR #125, opened 2022-09-23
Fixed ESI-Leap API WSGI script

samuel-zuk:
Previously, the ESI-Leap install process would create an executable script `esi-leap-wsgi-api` which did nothing and was broken. This pull request adds the missing functionality and allows ESI-Leap to be run properly via uWSGI and the like. The script was also renamed to `esi-leap-api-wsgi` from `esi-leap-wsgi-api` to better mirror the naming convention of other OpenStack services (and since the script was broken no one was using it anyway so this should not be an issue)

### DIFF
--- a/esi_leap/api/wsgi.py
+++ b/esi_leap/api/wsgi.py
@@ -10,10 +10,26 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-import oslo_i18n as i18n
+import sys
 
-_translators = i18n.TranslatorFactory(domain='esi-leap')
+from oslo_log import log as logging
 
-# The primary translation function using the well-known name "_"
-_ = _translators.primary
-install = i18n.install
+from esi_leap.api.app import WSGIApplication
+from esi_leap.common import i18n
+from esi_leap.common import service
+import esi_leap.conf
+
+
+CONF = esi_leap.conf.CONF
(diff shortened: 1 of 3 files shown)


esi_leap/api/wsgi.py
```diff
@@ -0,0 +1,44 @@
...
+from esi_leap.common import i18n
+from esi_leap.common import service
+import esi_leap.conf
+
+
+CONF = esi_leap.conf.CONF
+LOG = logging.getLogger(__name__)
+
+
+class EsiLeapApiWsgiApplication(object):
```
tzumainn, line 27: I wouldn't hate it if this was 'ESILeapAPIWSGIApplication - although staring at that string kinda makes my eyes hurt so maybe not. I noticed the Ironic version has much simpler naming conventions, so maybe that's okay... ?

DanNiESh: It looks like ironic has a similar class in app.py. Do we want to move this class to app.py?

samuel-zuk: hmm, i do think moving it to app.py would be a solid call and since it's already in the ESI-Leap repo and in the esi_leap.api namespace we could probably just call it "WSGIApplication" (or "WsgiApplication")

PR #126, opened 2022-11-18
Containerized esi-leap

rd4398:
This is a basic skeleton for the container file which can be used for containerized esi-leap. I referred to following links:

https://docs.docker.com/engine/reference/builder/
https://www.pybootcamp.com/blog/how-to-containerize-python-application/
https://people.redhat.com/~jupittma/cheatsheets/podman.html
https://www.pybootcamp.com/blog/how-to-write-dockerfile-python-apps/
https://snyk.io/blog/best-practices-containerizing-python-docker/

I need your inputs for further editing this specifically keeping esi-leap in mind

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,46 @@
+# This file will contain the configuration for containerized esi-leap
+
+# Use python as the parent image for esi-leap
+
+FROM python:3
+
+# Use label descriptor
+
+LABEL description="ESI-LEAP container file"
+
+# Maintainer details
+
+MAINTAINER <tzumainn@redhat.com>
+
+# The working directory for esi-leap inside the container
+
+WORKDIR /esi-leap
+
+# Copy the files of esi leap inside the container
+# Syntax : COPY <src> <destination>
+
+COPY . .
+
+# Run command to install dependencies for esi-leap
+
+RUN pip install esi-leap
+
+# Expose the default port for esi-leap-api
+# Use -p flag while running the image to map ports
+
+EXPOSE 7777
+
+USER root
+
+# Something for esi-leap-api and esi-leap-manager
+# Not sure how to link these two here / bring inside the container
+
+# Start processes for esi-leap-api and esi-leap-manager
+
+# What is the entry point / executable for esi-leap?
+
+RUN sudo esi-leap-dbsync create_schema
+
+RUN esi-leap-manager
+
+RUN esi-leap-api


rd4398:
Also, not sure why PR checks are failing...
I did not change any code, just added a new file named 'containerfile'!

Containerfile
```diff
@@ -0,0 +1,34 @@
...
+
+MAINTAINER <tzumainn@redhat.com>
+
+# The working directory for esi-leap inside the container
+
+WORKDIR /esi-leap
+
+# Copy the files of esi leap inside the container
+
+COPY . .
```
DanNiESh, line 21: Can you add the files that need to be copied to the container in this repo? I think esi-leap.conf should be included and saved under /etc/esi-leap.

Containerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN pip install esi-leap
+
+# Something for esi-leap-api and esi-leap-manager
+# Not sure how to link these two here / bring inside the container
+
+# Start processes for esi-leap-api and esi-leap-manager
+
+# What is the entry point / executable for esi-leap?
+
+CMD ["python", ./executable]
```
DanNiESh, line 34: Maybe use the commands here? https://github.com/CCI-MOC/esi-leap#run-the-services

Containerfile
```diff
@@ -0,0 +1,34 @@
...
+LABEL description="ESI-LEAP container file"
+
+# Maintainer details
+
+MAINTAINER <tzumainn@redhat.com>
+
+# The working directory for esi-leap inside the container
+
+WORKDIR /esi-leap
+
```
DanNiESh, line 18: You could add an "EXPOSE <port>" line in Dockerfile to expose esi-leap-api port

rd4398:
Okay, will do
Can you tell me on which port esi-leap-api runs?

DanNiESh:
> Okay, will do Can you tell me on which port esi-leap-api runs?

It's 7777 by default. https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/conf/api.py

rd4398:
Ah, okay!

rd4398:
The container build is failing for the command : "sudo esi-leap-dbsync create_schema" with error as "sudo command not found".... Any work around for this?
I guess we need this command to run inside the container....

DanNiESh:
> sudo command not found

Can you try to add "USER root" in dockerfile before "RUN sudo esi-leap-dbsync create_schema". I think you need to be root user to run esi-leap-xxx.

> The container build is failing for the command : "sudo esi-leap-dbsync create_schema" with error as "sudo command not found".... Any work around for this? I guess we need this command to run inside the container....

rd4398:
Okay, will try!

rd4398:
I am using '$COPY . .' in container file which copies entire esi-leap folder in the container, which is making me believe that esi-leap.conf will also be getting copied in the container. I am not able to figure out why I am still getting 'sudo command not found error' for "esi-leap-dbsync create schema". This needs to be figured out....

PR #129, opened 2023-01-19
Updated CI, fixed dependency issues

samuel-zuk:
This pull request updates this project's CI config and addresses three dependency issues:

- Kombu relied on some now deprecated `importlib_metadata` behavior. This behavior has been removed in the latest version of `importlib_metadata` and Kombu has released a patch to address this. In Python 3.8, `importlib.metadata` was pulled into the CPython standard library, meaning those versions never needed to install the external `importlib_metadata` package. But since we support Python 3.6 and 3.7, and the Kombu patch has not made it into the release version (on PyPi) yet. So until it does, we should pin `importlib_metadata` to versions under 5.0.0. This fixes previous problems with running unit tests on Python 3.7. (sidenote: Python 3.6 was not affected since the newest version of `importlib_metadata` is not available for it)
- An update to `dnspython` caused a breakage in `eventlet`-- this was fixed in the latest version of `eventlet` (0.33.3) and the package is now pinned to this version or higher to prevent future dependency conflicts. Closes #128 (check that PR for more background on this).
- ESI-Leap uses features incompatible with SQLAlchemy 2.0, which produces warnings telling us to pin `sqlalchemy` to versions before 2.0. This has been done and an environment variable has been set in `tox.ini` to silence the warning.

Additionally, as part of https://github.com/CCI-MOC/esi/issues/330, this project will now lint check and run unit tests during CI for versions of Python between 3.6, the earliest version that doesn't cause [other] major dependency breakages-- thru 3.11, the newest version currently available by default on the Ubuntu 20.04 CI runner. (Ubuntu 20.04 is itself, currently the newest version of Ubuntu with Python 3.6 installed by default)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,15 +5,15 @@ on: [push, pull_request]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
 
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


tzumainn (approved):
Looks good - thanks!

PR #132, opened 2023-02-21
dockerize-esi-leap

skanthed:


Containerfile
```diff
@@ -0,0 +1,55 @@
...
+
+# The working directory for esi-leap inside the container
+
+WORKDIR /esi-leap
+
+
+# Copy the files of esi leap inside the container
+# Syntax : COPY <src> <destination>
+
+COPY . .
```
tzumainn, line 23: It's generally better to be explicit regarding which files are actually needed; run_services.sh is one, there may be others.

Containerfile
```diff
@@ -0,0 +1,55 @@
...
+# Run command to install dependencies for esi-leap
+
+RUN pip install esi-leap
+
+RUN pip install mysql-connector
+
+# Please make sure to have esi-leap.conf file in same directory as containerfile 
+# or make sure to mention the right path of the file below.
+
+COPY esi-leap.conf /etc/esi-leap/esi-leap.conf
```
tzumainn, line 34: I'd move this COPY to be with the other COPY command

Containerfile
```diff
@@ -0,0 +1,55 @@
...
+
+EXPOSE 7777
+
+#USER root
+
+
+# Connect to database and Start processes for esi-leap-api and esi-leap-manager
+
+
+#RUN esi-leap-dbsync create_schema
```
tzumainn, line 51: I'm guessing this is commented out because the schema might already exist; however, that won't always be the case. Is it possible to have a script that creates the schema if it doesn't exist?

run_services.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Start the first process to connect to database 
+python3 connect.py
```
tzumainn, line 4: It's a bit strange to me that this is needed, since the esi-leap services should also just connect to the database. What behavior do you see if you take this out?

tzumainn (changes requested):
This is a good first iteration! I've left some comments inline; I'd also suggest creating a section in the README detailing what needs to be done to create a container.

Containerfile
skanthed: Yes. I made that change.

run_services.sh
skanthed: No, Actually that's for my testing purpose in an earlier phase. Yes, we don't need this I will remove it.

Containerfile
skanthed: Moved.

README.md
```diff
@@ -147,6 +147,22 @@ Once that's done, you can run the manager and API services:
     $ sudo esi-leap-api
 ```
 
+### Installation using Containerization
+
+By encapsulating the ESI Leap into a container, all the necessary dependencies, configurations and services are bundled into a single package.
```
tzumainn, line 152: It's probably worth mentioning that esi-leap.conf needs to be generated, and needs to be present in the directory that you're running `podman build` from.

skanthed: Updated readme and pushed the changes @tzumainn

tzumainn:
I tried testing this on the demo server; then I realized that the esi-leap endpoint and service hadn't been created. Can you try doing so following the instructions in https://github.com/CCI-MOC/esi-leap#create-the-openstack-service ? After that, you should be able to run 'openstack esi lease list' and similar commands. If your container is running on a different port, then you can change the endpoint created to use the container port.

skanthed:
@tzumainn I have updated the Readme of containerization section to follow the steps to create the OpenStack service and endpoint and also the Container file. Please look into it now.

Containerfile
```diff
@@ -0,0 +1,53 @@
...
+
+RUN pip install esi-leap
+
+RUN pip install mysql-connector
+
+RUN mkdir -p /var/log/esi-leap
+
+RUN touch /var/log/esi-leap/esi-leap-dbsync.log
+
+RUN apt update
```
DanNiESh, line 39: I think we use yum/dnf .rpm-based more commonly in ESI environment.

run_services.sh
samuel-zuk, line 1: This script will get the processes minimally up and running; however, I think this method of running these services contains a potentially severe issue. Lines 4 and 7 both spawn subprocesses before blocking on execution of `esi-leap-api`. When `esi-leap-api` exits, the script reaches its end and also exits, orphaning any subprocesses that are still running. Orphan processes become the responsibility of the init system, but since this was specified as the default command in the Containerfile, (I believe) it effectively *is* the init system.

From https://phusion.github.io/baseimage-docker/:

> Furthermore, `docker stop` sends SIGTERM to the init process, which is then supposed to stop all services. If your init process is your app, then it'll probably only shut down itself, not all the other processes in the container. The kernel will then forcefully kill those other processes, not giving them a chance to gracefully shut down, potentially resulting in file corruption, stale temporary files, etc.

This is potentially really bad, since `esi-leap-dbsync` and `esi-leap-manager` *both* make database queries. So if for example, there's a problem with the config that only affects the API, `esi-leap-api` will exit, and if either of the other two commands are still executing, they will be terminated immediately (I believe the kernel sends SIGKILL)-- potentially resulting in database corruption. In fact, I'm not sure if it's possible to ensure everything exits gracefully if the services are being run this way.

There are other issues to this approach as well: there's not really a good way of knowing if the manager process has went down other than looking at logs, and there's also not really a good way of restarting it without restarting the API as well. I'm personally not sure what the best course of action would be in resolving these issues, and I believe the solution likely warrants further discussion.

Containerfile
```diff
@@ -0,0 +1,51 @@
+# This file will contain the configuration for containerized esi-leap
+
+# Use python as the parent image for esi-leap
+
+FROM python:3.11
+
+# Use label descriptor
+
+LABEL description="ESI-LEAP container file"
```
samuel-zuk, line 9: I'd probably change "container file" to "container image" here, since `LABEL` describes the image being created.

samuel-zuk (changes requested):

run_services.sh
tzumainn: Could something like `trap` work here? https://unix.stackexchange.com/questions/259939/kill-a-running-shell-script-and-kill-all-process-running-within-it

samuel-zuk: It seems like the problem presented in that post is the inverse of ours-- they have a shell script running in the background that executes synchronous commands, whereas we have a shell script running in the foreground [of a container] that executes asynchronous commands. I think such a use of `trap` would allow us to make sure `docker stop` sends SIGTERM instead of SIGKILL to all child processes, but I don't see how it would allow us to ensure that the API process can exit without breaking things, as when that happens, the script will still finish execution and the kernel will still send SIGKILL to the child processes (importantly, SIGKILL cannot be trapped).

tzumainn: If `docker stop` sends SIGTERM, is there a reason why we can't trap the signal and run something that stops esi-leap-manager and esi-leap-api a bit more cleanly?

run_services.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "All processes stopped."
+    exit 0
+}
+
+# Register the stop_all_processes function to run on SIGTERM signal
+trap stop_all_processes SIGTERM
+
+# Start the first process
+esi-leap-dbsync create_schema &
+dbsync_pid=$!
```
samuel-zuk, line 17: This command should be ran in the foreground, not the background. Additionally, if the command is unsuccessful (e.g. exits with a non-zero status code) the script should exit without launching the API or the manager service.

Containerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN dnf install python3-pymysql
+
+RUN pip3 install pymysql
+
+# Expose the default port for esi-leap-api
+# Use -p flag while running the image to map ports
+
+EXPOSE 7777
+                                                                                                                                                                                                                  2,0-1
```
samuel-zuk, line 49: This line should should be removed. It will cause the build to fail.

Containerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN pip install mysql-connector
+
+RUN mkdir -p /var/log/esi-leap
+
+RUN touch /var/log/esi-leap/esi-leap-dbsync.log
+
+RUN dnf update -y
+
+RUN dnf install python3-pymysql
```
samuel-zuk, line 41: Do the Python container images ship with a Fedora-based base OS? I thought they were Debian-based (and if they are these lines should be replaced with equivalent `apt` commands).

Containerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Copy the files of esi leap inside the container
+# Syntax : COPY <src> <destination>
+
+COPY run_services.sh run_services.sh
+
+# Please make sure to have esi-leap.conf file in same directory as containerfile 
+# or make sure to mention the right path of the file below.
+
+#COPY esi-leap.conf /etc/esi-leap/esi-leap.conf
```
samuel-zuk, line 27: These lines should be removed from the Containerfile since they no longer apply to how we are currently doing things

README.md
```diff
@@ -147,6 +147,26 @@ Once that's done, you can run the manager and API services:
...
+
+Please make sure to follow the instructions in the [Configuration](#configuration) section to generate the esi-leap.conf file, and ensure that it is present in the directory from which you are running the `podman build` command.
```
samuel-zuk, line 154: Since we're no longer copying the configuration file into the container image during the build process itself, it is no longer necessary for the file to be in the working directory.

README.md
```diff
@@ -147,6 +147,26 @@ Once that's done, you can run the manager and API services:
...
+```
+    $ podman build -t <image-name> -f Containerfile . 
+```
+
+Once the container image is built, you can run it using the podman run command. The syntax is as follows:
+
+```
+    $ podman run --name <container-name> -p <host-port>:<container-port> -d -i -t -v <host-directory>:<container-directory> <image-name> /bin/sh
+```
+
```
samuel-zuk, line 169: This section should mention somewhere that the esi-leap.conf file must be mounted so that it's located at `/etc/esi-leap/esi-leap.conf` in the container. I'm also not sure what the purpose of the `-d -i -t` flags are in the sample command here-- they seem contradictory since `-d` is used to run the container in the background and `-i -t` are used to run the container in the current terminal. Finally, since there is a command to be run by default specified in the Containerfile, there should be no need to specify `/bin/sh` after `<image-name>`.

samuel-zuk (changes requested):
The core functionality looks good here, but a few further changes are required before I think this is ready to be merged.

tzumainn (approved):
Looks good!

Issue #131, opened 2023-02-08
Creating an offer or lease on a dummy node, then deleting the dummy node file causes all 'list' requests to fail.


samuel-zuk:
Encountered while running functional tests-- two tests were failing consistently and both involved "offer list." The OpenStack client was returning an odd error along the lines of:

`FileNotFoundError: [Errno 2] No such file or directory: '/tmp/nodes/<some_uuid>'`

The UUID in the error message was consistent every time, which didn't make sense considering the tests create dummy nodes with new random names each time. I got the same error running `openstack esi offer list` on the command line, even though other OpenStack commands worked fine. Turns out there was an offer created on a dummy node with the UUID '<some_uuid>' and its corresponding file in `/tmp/nodes` didn't exist. Since esi-leap apparently tries to read/locate these dummy node files when returning active offers/leases, it fails here in a place it shouldn't, with an unhelpful error message. 

It can be fixed, either by putting some valid JSON in the file esi-leap is looking for or by manually updating/deleting the offer/lease in the db, but this should still be fixed in esi-leap itself.

PR #133, opened 2023-04-25
Updated exception handling for resource objects

samuel-zuk:
Addresses #131.

Changes the internal behavior of the various `get_*` interfaces of dummy nodes and Ironic nodes to prevent uncaught internal exceptions from causing unexpected API errors.


tzumainn (approved):
Looks good - thanks!

PR #135, opened 2023-05-18
Resource object interface refactoring

samuel-zuk:
The interfaces for interacting with resource objects are inconsistent in naming and behavior. This patch seeks to address this by making the following changes:

- `get_resource_uuid()` and `get_resource_name()` have been renamed to `get_uuid()` and `get_name()`, respectively. This was done to make `Object.get_attribute()`-style calls more concise and legible.
- `get_node_config()` has been renamed to `get_config()`. This was done in part for the same reason as the previously mentioned change, but was also done to make it easier to potentially add new types of resource objects that aren't nodes in the future.
- `resource_admin_project_id()` has been renamed to `get_owner_project_id()`. The `get_` prefix was added for the sake of naming consistency and the term `admin` was replaced with `owner`. The latter change was made because in the existing code, the names of the attributes being retrieved contain the word `owner` instead of `admin`.
- `get_project_id()` has been renamed to `get_lessee_project_id()` to prevent confusion with the newly named `get_owner_project_id()`.
- `expire_lease()` has been renamed to `remove_lease()` to more accurately reflect what the method is doing. Specifically, in the existing code, the implementations of this function never expire the lease [in the ESI Leap database]-- they only remove the association between the lease/lessee and the resource object itself.
- The values returned by the various `get_` methods that *can* fail have been made consistent across the existing types of resource objects.
- The unit test suite for the resource object base has been updated to allow the base class to be tested as directly as possible.

### DIFF
--- a/esi_leap/api/controllers/v1/lease.py
+++ b/esi_leap/api/controllers/v1/lease.py
@@ -104,7 +104,7 @@ class LeasesController(rest.RestController):
                 resource_type = CONF.api.default_resource_type
 
             resource = get_resource_object(resource_type, resource_uuid)
-            resource_uuid = resource.get_resource_uuid()
+            resource_uuid = resource.get_uuid()
 
         filters = LeasesController._lease_get_all_authorize_filters(
             cdict, project_id=project_id, owner_id=owner_id,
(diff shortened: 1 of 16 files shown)


tzumainn (approved):
Looks super solid, and the changes make sense to me. Thanks!

PR #136, opened 2023-05-18
Fix various issues with esi-leap Containerfile

larsks:
This commit fixes several issues with the esi-leap Containerfile:

1. The `MAINTAINER` field is deprecated [1] and should not be used. The
   maintainer can change over time; using `git log` is a better method for
   determining who is responsible for a file.

2. Be consistent about using `pip3` or `pip` and consolidate the `pip3
   install` commands. This generally results in a shorter build time
   because we only need to calculate dependencies once.

3. Make the `CMD` entry independent of the permissions on
   `run_services.sh`.

[1]: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,44 +1,24 @@
-# This file will contain the configuration for containerized esi-leap
+# This file contains the configuration for containerized esi-leap
 
 # Use python as the parent image for esi-leap
-
 FROM python:3.11
 
-# Use label descriptor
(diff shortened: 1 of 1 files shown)


tzumainn (approved):
Ah, thanks for stepping in with your added expertise! I'll keep this in mind for the future.

PR #141, opened 2023-06-15
Add purpose field to lease notification

DanNiESh:
Need to be tested on demo server


tzumainn:
Tested, works, thanks!

tzumainn (approved):

PR #142, opened 2023-06-20
Add lease expire notification

DanNiESh:
Tested on demo server. Lease expire notification worked.

### DIFF
--- a/esi_leap/objects/lease.py
+++ b/esi_leap/objects/lease.py
@@ -208,25 +208,10 @@ class Lease(base.ESILEAPObject):
             LOG.info('Deleting lease %s', self.uuid)
             try:
                 resource = self.resource_object()
-                if resource.get_lease_uuid() == self.uuid:
-                    resource.remove_lease(self)
-                    if self.parent_lease_uuid is not None:
-                        parent_lease = Lease.get(self.parent_lease_uuid)
-                        resource.set_lease(parent_lease)
+                self.deactivate(context, resource)
+                self.status = statuses.DELETED
+                self.expire_time = datetime.datetime.now()
 
-                notify.emit_start_notification(context, self,
-                                               'delete', CRUD_NOTIFY_OBJ,
-                                               node=resource)
-                with notify.handle_error_notification(context,
-                                                      self,
-                                                      'delete',
-                                                      CRUD_NOTIFY_OBJ,
-                                                      node=resource):
-                    self.status = statuses.DELETED
-                    self.expire_time = datetime.datetime.now()
-                notify.emit_end_notification(context, self,
-                                             'delete', CRUD_NOTIFY_OBJ,
-                                             node=resource)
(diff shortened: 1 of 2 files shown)


esi_leap/objects/lease.py
```diff
@@ -302,8 +302,19 @@ def expire(self, context=None):
...
-                self.expire_time = datetime.datetime.now()
+                notify.emit_start_notification(context, self,
+                                               'delete', CRUD_NOTIFY_OBJ,
+                                               node=resource)
+                with notify.handle_error_notification(context,
+                                                      self,
+                                                      'delete',
+                                                      CRUD_NOTIFY_OBJ,
+                                                      node=resource):
+                    self.status = statuses.EXPIRED
```
tzumainn, line 313: Thinking about it, it probably makes the most sense for this `with` block to surround lines 298-303 above, as that's what actually unsets the ironic lessee field. I might consider:

* making this change and also isolating it into its own function, so that here we just call that function and then set the status to EXPIRED and set the expire_time here
* update `cancel` to use this new function
* update `fulfill` to have the `with` block surround the call to `set_lease`

tzumainn (changes requested):

esi_leap/objects/lease.py
DanNiESh: Will do!

DanNiESh:
New changes updated. Tested in demo server.

esi_leap/objects/lease.py
```diff
@@ -208,25 +208,10 @@ def cancel(self, context=None):
             LOG.info('Deleting lease %s', self.uuid)
             try:
                 resource = self.resource_object()
-                if resource.get_lease_uuid() == self.uuid:
-                    resource.remove_lease(self)
-                    if self.parent_lease_uuid is not None:
-                        parent_lease = Lease.get(self.parent_lease_uuid)
-                        resource.set_lease(parent_lease)
+                self.emit_delete_notification(context, resource)
```
tzumainn, line 211: I'd rename this function, since its main purpose isn't to emit the notification; it's to actually unset the lease, with the notification being a side effect. Maybe something like `deactivate`?

esi_leap/objects/lease.py
```diff
@@ -263,12 +245,19 @@ def fulfill(self, context=None):
                                                       'fulfill',
                                                       CRUD_NOTIFY_OBJ,
                                                       node=resource):
-                    self.status = statuses.ACTIVE
-                    self.fulfill_time = datetime.datetime.now()
+                    try:
+                        resource.set_lease(self)
+                    except Exception:
+                        raise
```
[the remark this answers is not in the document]
DanNiESh: Oh, yes. It raised the exception. I removed the try exception block.

tzumainn (approved):

tzumainn:
Looks good - thanks!

PR #144, opened 2023-06-28
Added event object

tzumainn:
The event object retains a record of all events. The available filters are designed to allow an end user to view all events past the last time the events were queried.

### DIFF
--- a/esi_leap/api/controllers/v1/event.py
+++ b/esi_leap/api/controllers/v1/event.py
@@ -1,0 +1,113 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import datetime
+import pecan
+from pecan import rest
+import wsme
+from wsme import types as wtypes
+import wsmeext.pecan as wsme_pecan
+
+from esi_leap.api.controllers import base
+from esi_leap.api.controllers import types
+from esi_leap.api.controllers.v1 import utils
+from esi_leap.common import exception
+from esi_leap.common import keystone
+import esi_leap.conf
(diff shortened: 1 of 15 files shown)


esi_leap/api/controllers/v1/event.py
```diff
@@ -0,0 +1,113 @@
...
+                         datetime.datetime, wtypes.text, wtypes.text,
+                         wtypes.text, wtypes.text)
+    def get_all(self, last_event_id=None, lessee_or_owner_id=None,
+                last_event_time=None, event_type=None,
+                resource_type=None, resource_uuid=None):
+        request = pecan.request.context
+        cdict = request.to_policy_values()
+
+        try:
+            utils.policy_authorize('esi_leap:offer:offer_admin', cdict, cdict)
```
DanNiESh, line 70: I'm a little bit confused: why it checks policy esi_leap:offer:offer_admin? And who can make call to event get_all()? Admin, lessee and owner?

tzumainn: Ah, right - basically, anyone can make a request. However, if the person requesting a list of events is *not* an admin, then they must be restricted to themselves. So this snippet says: if the user is not an admin, then force the query to be limited to their own project; otherwise, it's okay for the user to specify whatever project they want

DanNiESh (commented):
I have one question about the event object. When users list events, they are actually checking lease fulfill and expire/delete report. Do we want a real-time system that the users can react to events immediately as they occur?

tzumainn:
> I have one question about the event object. When users list events, they are actually checking lease fulfill and expire/delete report. Do we want a real-time system that the users can react to events immediately as they occur?

Fair question! The answer is - that's a lot more difficult. This system will enable users to periodically query for new events, using the `last_notification_time` or `last_event_id` filters to make sure there's no duplicates. Examples for that will be forthcoming.

esi_leap/db/sqlalchemy/models.py
```diff
@@ -112,3 +112,25 @@ class Lease(Base):
...
         backref=orm.backref('child_leases', remote_side=uuid),
     )
+
+
+class Event(Base):
+    """Represents an event."""
+
+    __tablename__ = 'events'
+    __table_args__ = (
+        Index('event_type_idx', 'event_type'),
```
DanNiESh, line 122: I think event_time could also be an index since searching by last_event_time might be a common use case.

tzumainn: you're definitely right - good catch! I've added it

DanNiESh (approved):
LGTM

PR #145, opened 2023-07-13
Fix bug that causes expired leases to not be reflected on ironic nodes

tzumainn:
The ironic_node resource_object automatically popped the lease_uuid from its properties dict when calling get_config. However, this is permanently reflected on the resource object, and results in future queries of the resource object's lease_uuid returning None.


DanNiESh (approved):
LGTM

PR #147, opened 2023-10-20
Add a max lease length and a default length

DanNiESh:


esi_leap/api/controllers/v1/lease.py
```diff
@@ -156,6 +156,9 @@ def post(self, new_lease):
                                        lease_dict['resource_uuid'])
         lease_dict['resource_uuid'] = resource.get_uuid()
 
+        max_lease_time = CONF.api.max_lease_time
+        default_lease_time = CONF.api.default_lease_time
```
tzumainn, line 160: I think it may be cleaner to not have these explicit assignments, and simply call `CONF.api.*_lease_time` at the appropriate points in the code.

tzumainn (changes requested):
One minor comment; looks great otherwise!

esi_leap/api/controllers/v1/lease.py
DanNiESh: Fixed!

tzumainn (approved):

tzumainn:
Looks good - thanks!

PR #148, opened 2023-11-01
Add email notification when lease expires soon

DanNiESh:


esi_leap/common/keystone.py
```diff
@@ -73,3 +73,11 @@ def get_project_name(project_id, project_list=None):
         return project.name if project else ''
     else:
         return ''
+
+
+def get_project_email(project_id):
+    if project_id:
+        project = get_keystone_client().projects.get(project_id)
```
tzumainn, line 80: Will this return 'None' if project_id is 'None'? If so, the first if statement may not be needed

esi_leap/common/keystone.py
```diff
@@ -73,3 +73,11 @@ def get_project_name(project_id, project_list=None):
...
+        return project.email if project else ''
```
tzumainn, line 81: What if a project doesn't have an email attribute? Will this still work?

esi_leap/common/keystone.py
```diff
@@ -73,3 +73,11 @@ def get_project_name(project_id, project_list=None):
...
+    else:
+        return ''
```
tzumainn, line 83: It may make sense to make the default 'None' instead of an empty string

esi_leap/conf/api.py
```diff
@@ -23,6 +23,8 @@
     cfg.StrOpt('default_resource_type', default='ironic_node'),
     cfg.IntOpt('max_lease_time', default=21),
     cfg.IntOpt('default_lease_time', default=7),
+    cfg.StrOpt('SMTP_server', default='localhost'),
+    cfg.StrOpt('email_sender', default='admin@esi.massopen.cloud'),
```
tzumainn, line 27: I don't think we want a default to be MOC-specific; we maybe shouldn't have a default for this at all.

It may make sense to include options for

* whether emails should be sent at all
* for the "lease expires in X days", a configuration for X

esi_leap/manager/service.py
```diff
@@ -130,6 +140,62 @@ def _expire_offers(self):
...
+            if lease.end_time - now <= datetime.timedelta(1):
+                try:
+                    LOG.info('Lease %s is expiring in 1 day', lease.uuid)
+                    self._send_email(lease)
+                except Exception as e:
+                    LOG.info('Error sending lease expire '
+                             'email notification: %s: %s' %
+                             (type(e).__name__, e))
+
+    def _send_email(self, lease):
```
tzumainn, line 160: I'd put this into its own utility function which takes in the 'to_email' and the body of the email

esi_leap/manager/service.py
```diff
@@ -130,6 +140,62 @@ def _expire_offers(self):
...
 
+    def _notify_lease_expire(self):
+        LOG.info('Checking for leases expire soon')
+        leases = lease_obj.Lease.get_all(
+            {'status': [statuses.ACTIVE, statuses.CREATED,
+                        statuses.WAIT_EXPIRE, statuses.WAIT_FULFILL]},
+            self._context)
+        now = timeutils.utcnow()
+        for lease in leases:
+            if lease.end_time - now <= datetime.timedelta(1):
```
tzumainn, line 151: If this runs every 6 hours, won't a user get up to four emails notifying them that their lease is about to expire?

tzumainn (changes requested):
Hi! This is a good start; some comments inline.

esi_leap/common/keystone.py
DanNiESh: I have simplified the code here

esi_leap/common/keystone.py
DanNiESh: Oh, yes. I changed it to getattr(project, 'email', None) to get email value or None

esi_leap/common/keystone.py
DanNiESh: Fixed

esi_leap/conf/api.py
DanNiESh: Yep! I removed the email_sender default value and add lease_warning_days and enable_email configurations

esi_leap/manager/service.py
DanNiESh: done

esi_leap/manager/service.py
DanNiESh: I changed it to 24 hours, so users will get at most one email for one day

tzumainn:
@DanNiESh should we close this?

DanNiESh:
> @DanNiESh should we close this?

Yep. I'll submit a PR of the updated script to esi-pilot

Issue #149, opened 2023-10-03
Create email notifications when a lease is close to expiration


tzumainn:
Notify users (projects?) when their lease is about to go away so they don't panic.

joachimweyl:
@DanNiESh do we plan to finish this in this sprint or shall we push it off to the next sprint?

DanNiESh:
Please push it off to the next sprint. This task should be done quickly

DanNiESh:
https://github.com/CCI-MOC/esi-leap/pull/150

PR #150, opened 2023-12-11
Send email notification when lease starts or ends soon

DanNiESh:
Closes: #149


larsks:
You should update your commit message and the PR description to include a link to the issue with which it is associated. This will update the issue to include a link to the commit (for mention in the commit message) and to this PR (for the mention in the PR description).

It's generally good practice for your commit message to provide some context for the changes.

If you write something like:

```
Closes: https://github.com/CCI-MOC/esi-leap/issues/149
```

Then the referenced issue will be closed automatically when this PR merges.

DanNiESh:
> You should update your commit message and the PR description to include a link to the issue with which it is associated. This will update the issue to include a link to the commit (for mention in the commit message) and to this PR (for the mention in the PR description).
> 
> It's generally good practice for your commit message to provide some context for the changes.
> 
> If you write something like:
> 
> ```
> Closes: https://github.com/CCI-MOC/esi-leap/issues/149
> ```
> 
> Then the referenced issue will be closed automatically when this PR merges. If the PR and the issue are in the same repository, you can write:
> 
> ```
> Closes: #149
> ```

Thank you for your advice! I've updated PR description as well as commit message with a link to the issue.

tzumainn:
@DanNiESh this looks pretty good - I think optimally we don't run openstack commands on the command line, but I think that can be updated later once the ESI commands get an SDK.

Can you add some comments to the top of the file explaining how to run the script?

DanNiESh:
> @DanNiESh this looks pretty good - I think optimally we don't run openstack commands on the command line, but I think that can be updated later once the ESI commands get an SDK.
> 
> Can you add some comments to the top of the file explaining how to run the script?

Done!

bin/query_events.py
```diff
@@ -0,0 +1,126 @@
...
+            node_uuid = event['Resource UUID']
+            lease_uuid = event['Object UUID']
+            lease = run_openstack_command('openstack esi lease show %s' % lease_uuid)
+            print("Lease %s with purpose %s on node %s started" % (lease_uuid, lease['purpose'], node_uuid))
+            if enable_email:
+                email_body_lease_start = f"Hi {lease['project']}," \
+                                         f"\n\n" \
+                                         f"We would like to inform you that " \
+                                         f"your lease: {lease['uuid']} is fulfilled. " \
+                                         f"Start Time: {lease['start_time']}." \
```
tzumainn, line 88: Sorry, I should have checked the text earlier; could you change "is fulfilled" to "has started", and also include the End Time and the node name and UUID?

bin/query_events.py
```diff
@@ -0,0 +1,126 @@
...
+    now = datetime.datetime.utcnow()
+    for lease in leases:
+        lease_end = datetime.datetime.strptime(lease["End Time"], "%Y-%m-%dT%H:%M:%S.%f")
+        if lease_end - now <= datetime.timedelta(lease_warning_days):
+            print('Lease %s is expiring in %s day(s)' %(lease["UUID"], lease_warning_days))
+            if enable_email:
+                email_body_lease_expire = f"Hi {lease['Project']}," \
+                                          f"\n\n" \
+                                          f"We would like to inform you that" \
+                                          f" your lease: {lease['UUID']} " \
```
tzumainn, line 107: Can we add the node UUID and name as well?

tzumainn (changes requested):
Just a few minor text changes!

bin/query_events.py
```diff
@@ -0,0 +1,126 @@
...
+    new_last_event_id = last_event_id
+    for event in events:
+        new_last_event_id = event['ID']
+        if event['Event Type'] == 'esi_leap.lease.fulfill.end':
+            node_uuid = event['Resource UUID']
+            lease_uuid = event['Object UUID']
+            lease = run_openstack_command('openstack esi lease show %s' % lease_uuid)
+            print("Lease %s with purpose %s on node %s started" % (lease_uuid, lease['purpose'], node_uuid))
+            if enable_email:
+                email_body_lease_start = f"Hi {lease['project']}," \
```
larsks, line 84: Should this content (and similarly elsewhere in the code) come from an external template instead (so that an administrator can change the text without needing to modify the code)?

DanNiESh:
I've updated the content of email body and provided external email templates for the script.

DanNiESh:
updated the endpoint name to esi-leap-email-notification

esi_leap/send_email_notification.py
```diff
@@ -0,0 +1,173 @@
...
+- SMTP_SERVER: Hostname of the SMTP server for sending emails. Default is
+  'localhost'
+- EMAIL_SENDER: The email address used as the sender in outgoing emails
+- LEASE_WARNING_DAYS: Days before a lease's end to send a notification email
+- LEASE_CREATE_TEMPLATE: Path to the email template for lease creation
+  notifications. Default is esi_leap/templates/lease_create_email.txt
+- LEASE_EXPIRE_TEMPLATE: Path to the email template for lease expiration
+  warnings. Default is esi_leap/templates/lease_expire_email.txt
+
+After esi-leap package is installed, run `esi_leap_query_events`
```
tzumainn, line 30: Would you run `esi-leap-email-notification` instead? I might mention that the script would be run periodically as well.

DanNiESh: The script itself doesn't have scheduling mechanism. But we can make it a cronjob to run periodically. Just add the command `esi-leap-email-notification` in a crontab file and specify a frequency would work.

tzumainn: Yep, sorry, I was unclear - the documentation should be updated to say "run esi-leap-email-notification" instead of "esi_leap_query_events", right? And I think the documentation should tell the operator to set up a periodic cron job rather than just running it once.

tzumainn (approved):

tzumainn:
Looks good - thanks!

esi_leap/send_email_notification.py
DanNiESh: Makes sense! I just added this instruction in the documentation.

PR #151, opened 2024-02-13
Fix bug with claiming offer

tzumainn:
If an offer is claimed for its entire duration, and someone else tries to claim the offer without specifying a start/end time, then the db function will think that the next available end time is before the current time. This fixes that issue, as well as renaming the db function to be clearer regarding its function.


tzumainn:
@DanNiESh could you take a look at this when you have a moment? Thanks!

PR #152, opened 2024-03-20
Push up node resource_class/properties/traits to node/lease/offer APIs

tzumainn:
These attributes are of interest to users, and should be viewable by users. This PR also renames resource_object 'config' to 'properties', as that is what that attribute actually represents.


DanNiESh (approved):
Looks good to me!

PR #153, opened 2024-04-02
Allow leases to update end_time

tzumainn:
Only end_time is allowed for now

### DIFF
--- a/esi_leap/objects/lease.py
+++ b/esi_leap/objects/lease.py
@@ -160,47 +160,48 @@ class Lease(base.ESILEAPObject):
 
     def create(self, context=None):
         updates = self.obj_get_changes()
-
-        with utils.lock(utils.get_resource_lock_name(updates['resource_type'],
-                                                     updates['resource_uuid']),
+        resource_type = updates['resource_type']
+        resource_uuid = updates['resource_uuid']
+        start_time = updates['start_time']
+        end_time = updates['end_time']
+        with utils.lock(utils.get_resource_lock_name(resource_type,
+                                                     resource_uuid),
                         external=True):
-            if updates['start_time'] >= updates['end_time']:
-                raise exception.InvalidTimeRange(
-                    resource='lease',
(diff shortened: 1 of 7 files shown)


tzumainn:
@DanNiESh would you be able to take a look at this when you have a moment? There's a matching CLI update as well.

esi_leap/objects/lease.py
```diff
@@ -160,47 +160,48 @@ def get_all(cls, filters, context=None):
...
-                parent_lease.verify_child_availability(updates['start_time'],
-                                                       updates['end_time'])
-            else:
-                ro = get_resource_object(updates['resource_type'],
-                                         updates['resource_uuid'])
-                ro.verify_availability(updates['start_time'],
-                                       updates['end_time'])
+            # only need to check availabilities if new end time is greater
+            # than previous end time
+            if new_end_time > self.end_time:
```
DanNiESh, line 196: In verify_time_range(), it also checks if new_end_time > self.end_time. Is it a duplicate check here?

tzumainn: Ah, the difference is that this check is to see if we even need to call `verify_time_range`.  Let's say lease initially runs from time `A----->B`. If the updated end time is before `B`, there's no need to `verify_time_range` at all, since the new lease time range is within the old one. If it's after time `B`, then we verify from `B---->C`.

DanNiESh: Yes, but if the updated end time is before `B`, do you need to raise an exception to indicate that the new_end_time is invalid?

tzumainn: Ah, nope - you're right, in the update case I technically don't need that check, since it'll never happen. However when I consolidated the time checking code into `verify_time_range`, it seemed strange to leave that out; I figured it was better to have a duplicated check than to leave out a condition if/when some other code calls `verify_time_range` in the future.

DanNiESh (approved):
Looks good!

PR #157, opened 2024-06-21
Ensured all necessary dependencies, including mysql-connector and pymysql, are listed in requirements.txt.

skanthed:


Containerfile
larsks, line 1: You need to call `pip install -r requirements.txt` somewhere. Currently, this image is broken -- the programs called in `run_services.sh` don't exist.

larsks (changes requested):

Containerfile
skanthed: Yes, I was testing it and was just about to update that change. That's the reason I marked this as a draft pull request. Updating it.

Containerfile
```diff
@@ -10,13 +10,13 @@ WORKDIR /esi-leap
...
 RUN touch /var/log/esi-leap/esi-leap-dbsync.log
 
-# Run command to install dependencies for esi-leap
-RUN apt update -y && apt install -y python3-pymysql
-RUN pip3 install esi-leap mysql-connector pymysql
+# Install dependencies
+RUN pip install -r requirements.txt
 
 # Indicate that service listens on this port
 EXPOSE 7777
```
larsks, line 22: You want to keep your `RUN pip install` operation as close as possible to the `COPY requirements.txt` operation: whenever you update `requirements.txt`, this invalidates the layer cache for every operation *after* the `COPY`. This means your container engine will need to re-run those steps as part of the build process rather than being able to utilize the cached layers, which increases the time required to build the image.

More broadly, to make the most effective use of the layer cache you want to order your operations from "least likely to change" to "most likely to change".

I might structure the file like this:

```
FROM docker.io/python:3.11

LABEL description="ESI-LEAP -- ESI lease policy manager"

# The working directory for esi-leap inside the container
WORKDIR /esi-leap

# Install dependencies
COPY requirements.txt requirements.txt
RUN pip install -r requirements.txt

RUN mkdir -p /var/log/esi-leap
RUN touch /var/log/esi-leap/esi-leap-dbsync.log

# Indicate that service listens on this port
EXPOSE 7777

# Copy the files of esi leap inside the container
COPY run_services.sh run_services.sh

CMD ["bash", "./run_services.sh"]
```

You can find more information about the Docker build cache [here](https://docs.docker.com/build/cache/). Podman (and buildah) operate on similar principles.

Containerfile
larsks, line 1: Note that when using Podman, you should get used to using fully qualified image names. Instead of:

```
FROM python:3.11
```

You should use:

```
FROM docker.io/python:3.11
```

When using unqualified names, depending on someone's local configuration, `python:3.11` may not refer to the same image as it does on your system.

Containerfile
skanthed: Understood and will take care of it from next time.

larsks (approved):

Containerfile
```diff
@@ -1,24 +1,29 @@
...
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
 
+# Create necessary directories and files
 RUN mkdir -p /var/log/esi-leap
 RUN touch /var/log/esi-leap/esi-leap-dbsync.log
 
-# Run command to install dependencies for esi-leap
-RUN apt update -y && apt install -y python3-pymysql
-RUN pip3 install esi-leap mysql-connector pymysql
```
larsks, line 19: Because you are no longer running `pip install esi-leap`, the resulting image does not actually have `esi-leap` installed. Attempting to run the image results in:

```
./run_services.sh: line 16: esi-leap-dbsync: command not found
```

You should ensure the image runs correctly (before submitting the pull request).

There are two ways of resolving this:

1. You could just re-introduce `pip install esi-leap`, but that is problematic because the version of esi-leap installed in the image won't match the version of esi-leap checked out in the local directory. This means the Containerfile wouldn't be useful for automatically building images on commits to `main`, for example.

2. Update the Containerfile to copy in the contents of the local directory and then `pip install .`.

I strongly favor option 2. @tzumainn?

larsks (changes requested):
It looks like there is a problem with the image; attempting to run it reports:

```

```

Containerfile
```diff
@@ -1,24 +1,29 @@
...
+EXPOSE 7777
+
 # The working directory for esi-leap inside the container
 WORKDIR /esi-leap
 
-# Copy the files of esi leap inside the container
-COPY run_services.sh run_services.sh
+# Copy the requirements file and install dependencies
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
```
larsks, line 16: Re: my earlier comment, that would look like this:

```suggestion
COPY requirements.txt requirements.txt
RUN pip install -r requirements.txt
COPY . ./
RUN pip install .
```

Containerfile
skanthed: I already made that change in the file I tested. I think during commit I forgot to do git add but the changes are already made and I tested it for sure. I will just push the latest commit. Sorry for this silly mistake.

larsks (approved):

tzumainn (approved):
Looks good - thanks!

joachimweyl:
@skanthed is this ready to merege?

tzumainn:
Whoops - thought I merged this.

PR #161, opened 2024-06-27
Formatting and content updates for README.md

larsks:
- Consistent use of Markdown headings: children of an h1 heading should be
  h2 headings
- Consistent line wrapping: previously, some paragraphs were line-wrapped
  and others were not. This commit makes all paragraphs unwrapped.
- Use `pip` for installation rather than `python setup.py`
- Make it easier to copy-and-paste most command examples by removing `$`
  shell prompts
- Remove unnecessary `GRANT` commands from mysql instructions: `%` is a wildcard so `'esi_leap'@'%'` already grants access to `'esi_leap'@'<database_ip>'`
- Simplify the "Container Installation" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,68 +1,54 @@
 # esi-leap
 
-esi-leap is an OpenStack service for leasing baremetal nodes, designed to run
-on top of [multi-tenant
-Ironic](https://docs.openstack.org/ironic/latest/admin/node-multitenancy.html).
-It consists of an API that provides endpoints for leasing operations and a
-manager service that updates the status of leases and offers as required. See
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,68 +1,54 @@
...
+pip install esi-leap
 ```
 
+To install from source:
 
-### Client
+```
+git clone https://github.com/CCI-MOC/esi-leap
+cd esi-leap
+pip install .
```
tzumainn, line 18: I have to admit, I didn't realize you could do this.

tzumainn (approved):
I remember vaguely thinking that the '##' heading looked too big, but looking at it now it looks fine. Thanks!

PR #162, opened 2024-07-10
Added dummy identity provider to remove Keystone dependancy during testing

### LINKED ISSUE #159: Allow developers to run esi-leap without keystone
It would be incredibly convenient for development purposes if it were possible to run the esi-leap service without requiring a keystone instance. Unfortunately, the existing implementation is strongly coupled to keystone. Solving this dependency issue is trickier than it needs to be due to some previous design decisions:

1. `esi_leap/api/controllers/v1/utils.py` has a bunch of standalone functions that should be instance methods.

    - `offer_get_dict_with_added_info` should be part of `OffersController`.

    - `lease_get_dict_with_added_info` should be part of `LeasesController`

    Because these functions are not part of their respective class instance, it is effectively impossible to modify their behavior through configuration at runtime.

2. Several of the existing controllers make calls directly to the `esp_leap.common.keystone` module.

The path forward is to replace this strong coupling with [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection). In particular, the controller classes should receive an IDP implementation object when they are instantiated, and use that for obtaining necessary metadata. That is, instead of call `keystone.get_project_list()`, we should have a base IDP interface like this:

```
class BaseIDP(abc.ABC):
  def get_project_list():
    ...

  def get_project_name(self, id, project_list=None):
    ...

  def get_project_uuid(project_ident):
    ...
```

Then we should implement a `KeystoneIDP` class that implements this interface; for normal production use we would pass that into the controllers.

For development use, we would have something like a `StaticIDP` that would take a project id, name, etc from a section in the `esi-leap.conf` file.

In addition to make development much more convenient, this also makes testing easier: it is no longer necessary to mock out keystone-related functions; we can instead just pass in a test-specific IDP class that always returns known values.

QuanMPhm:
Closes #159 after the draft is complete. The dummy identity provider (idp) can be enabled by setting the environment variable ESI_DEBUG to True. For now, the dummy idp returns information about a dummy project.

Some functions from `api/controllers/v1/utils.py` have been moved into the only controllers that use them and turned into static class methods.

Assuming everyone is fine with these draft changes, the remaining steps would be to do some cleanup with the test cases.

### DIFF
--- a/esi_leap/common/idp/__init__.py
+++ b/esi_leap/common/idp/__init__.py
@@ -1,0 +1,8 @@
+import os
+
+from esi_leap.common.idp import keystoneIDP, dummyIDP
+
+if os.environ.get('ESI_DEBUG', '') == 'True':
+    idp = dummyIDP.DummyIDP()
+else:
+    idp = keystoneIDP.KeystoneIDP()
--- a/esi_leap/common/idp/baseIDP.py
+++ b/esi_leap/common/idp/baseIDP.py
@@ -1,0 +1,15 @@
+import abc
+
+class BaseIDP(abc.ABC):
+  
+  def get_project_list():
+    pass
+
+  def get_project_name(self, id, project_list=None):
+    pass
+  
+  def get_parent_project_id_tree(project_id):
+    pass
+
+  def get_project_uuid_from_ident(project_ident):
+    pass
--- a/esi_leap/common/idp/dummyIDP.py
+++ b/esi_leap/common/idp/dummyIDP.py
@@ -1,0 +1,36 @@
+from oslo_utils import uuidutils
+from keystoneclient.v3.projects import Project, ProjectManager
+
+from esi_leap.common import exception
+from esi_leap.common.idp import baseIDP
+
+dummy_project = Project(manager=ProjectManager, info={
+    'id' : 1,
+    'name' : 'test'
+})
+
+class DummyIDP(baseIDP.BaseIDP):
+
+    def get_parent_project_id_tree(self, project_id):
+        return [1]
+
+
+    def get_project_uuid_from_ident(self, project_ident):
+        if uuidutils.is_uuid_like(project_ident):
+            return project_ident
+        else:
+            if project_ident == "test":
+                return 1
+            raise exception.ProjectNoSuchName(name=project_ident)
+
+
+    def get_project_list(self):
+
+        return [dummy_project]
+
+
+    def get_project_name(self, project_id, project_list=None):
+        if project_id == 1:
+            return 'test'
+        else:
+            return ''
--- a/esi_leap/api/controllers/v1/event.py
+++ b/esi_leap/api/controllers/v1/event.py
@@ -21,7 +21,7 @@ from esi_leap.api.controllers import base
 from esi_leap.api.controllers import types
 from esi_leap.api.controllers.v1 import utils
 from esi_leap.common import exception
-from esi_leap.common import keystone
+from esi_leap.common.idp import idp
 import esi_leap.conf
 from esi_leap.objects import event as event_obj
 from esi_leap.resource_objects import get_resource_object
@@ -72,7 +72,7 @@ class EventsController(rest.RestController):
             lessee_or_owner_id = cdict['project_id']
 
         if lessee_or_owner_id is not None:
-            lessee_or_owner_id = keystone.get_project_uuid_from_ident(
+            lessee_or_owner_id = idp.get_project_uuid_from_ident(
                 lessee_or_owner_id)
 
         if resource_uuid is not None:
--- a/esi_leap/api/controllers/v1/lease.py
+++ b/esi_leap/api/controllers/v1/lease.py
@@ -26,7 +26,7 @@ from esi_leap.api.controllers.v1 import utils
 from esi_leap.common import constants
 from esi_leap.common import exception
 from esi_leap.common import ironic
-from esi_leap.common import keystone
+from esi_leap.common.idp import idp
 from esi_leap.common import statuses
 import esi_leap.conf
 from esi_leap.objects import lease as lease_obj
@@ -84,7 +84,7 @@ class LeasesController(rest.RestController):
         lease = utils.check_lease_policy_and_retrieve(
             request, 'esi_leap:lease:get', lease_id)
 
-        return Lease(**utils.lease_get_dict_with_added_info(lease))
+        return Lease(**self._lease_get_dict_with_added_info(lease))
 
     @wsme_pecan.wsexpose(LeaseCollection, wtypes.text,
                          datetime.datetime, datetime.datetime, wtypes.text,
@@ -97,10 +97,10 @@ class LeasesController(rest.RestController):
         cdict = request.to_policy_values()
 
         if project_id is not None:
-            project_id = keystone.get_project_uuid_from_ident(project_id)
+            project_id = idp.get_project_uuid_from_ident(project_id)
 
         if owner_id is not None:
-            owner_id = keystone.get_project_uuid_from_ident(owner_id)
+            owner_id = idp.get_project_uuid_from_ident(owner_id)
 
         if resource_uuid is not None:
             if resource_type is None:
@@ -126,12 +126,12 @@ class LeasesController(rest.RestController):
 
             with concurrent.futures.ThreadPoolExecutor() as executor:
                 f1 = executor.submit(ironic.get_node_list)
-                f2 = executor.submit(keystone.get_project_list)
+                f2 = executor.submit(idp.get_project_list)
                 node_list = f1.result()
                 project_list = f2.result()
 
             leases_with_added_info = [
-                Lease(**utils.lease_get_dict_with_added_info(l, project_list,
+                Lease(**self._lease_get_dict_with_added_info(l, project_list,
                                                              node_list))
                 for l in leases]
             if resource_class:
@@ -159,7 +159,7 @@ class LeasesController(rest.RestController):
         lease_dict['resource_uuid'] = resource.get_uuid()
 
         if 'project_id' in lease_dict:
-            lease_dict['project_id'] = keystone.get_project_uuid_from_ident(
+            lease_dict['project_id'] = idp.get_project_uuid_from_ident(
                 lease_dict['project_id'])
 
         if 'start_time' not in lease_dict:
@@ -189,7 +189,7 @@ class LeasesController(rest.RestController):
 
         lease = lease_obj.Lease(**lease_dict)
         lease.create(request)
-        return Lease(**utils.lease_get_dict_with_added_info(lease))
+        return Lease(**self._lease_get_dict_with_added_info(lease))
 
     @wsme_pecan.wsexpose(Lease, wtypes.text, body={wtypes.text: wtypes.text})
     def patch(self, lease_uuid, patch=None):
@@ -208,7 +208,7 @@ class LeasesController(rest.RestController):
         updates = {'end_time': new_end_time}
         lease.update(updates, request)
 
-        return Lease(**utils.lease_get_dict_with_added_info(lease))
+        return Lease(**self._lease_get_dict_with_added_info(lease))
 
     @wsme_pecan.wsexpose(Lease, wtypes.text)
     def delete(self, lease_id):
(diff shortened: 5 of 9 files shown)


esi_leap/common/idp/__init__.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+from esi_leap.common.idp import keystoneIDP, dummyIDP
+
+if os.environ.get('ESI_DEBUG', '') == 'True':
```
larsks, line 5: If we're going to use an environment variable in the short term, I think it needs to be something more specific than `ESI_DEBUG` (because (a) this isn't really a debugging facility, and (b) `ESI_DEBUG` seems a bit broad and it would be tempting to make it do *other* things related to debugging as well). Maybe `ESI_USE_DUMMY_IDP` instead.

I think for the final version of this PR we want to introduce a new config option to control IDP selection, and it should allow us to provide an arbitrary fully qualified class name. Something like:

```
[esi-leap]
idp_plugin=esi_leap.common.idp.keystoneIDP
```
For testing, we should be able to pass in an arbitrary class instance that supports the necessary idp API. This would allow us to perform failure injection and other sorts of tests without needing to monkey patch the existing code.

esi_leap/common/idp/baseIDP.py
```diff
@@ -0,0 +1,15 @@
+import abc
+
+class BaseIDP(abc.ABC):
+  
+  def get_project_list():
```
larsks, line 5: If this is an `abc.ABC`, then we should be decorating these methods with the [`abstractmethod` decorator](https://docs.python.org/3/library/abc.html#abc.abstractmethod).

esi_leap/common/idp/dummyIDP.py
```diff
@@ -0,0 +1,36 @@
+from oslo_utils import uuidutils
+from keystoneclient.v3.projects import Project, ProjectManager
+
+from esi_leap.common import exception
+from esi_leap.common.idp import baseIDP
+
+dummy_project = Project(manager=ProjectManager, info={
+    'id' : 1,
```
larsks, line 8: Ids should be valid UUIDs; some functions validate their input and will throw warnings or errors if passed a non-UUID value.

larsks: Ideally, we should be able to provision the dummy data externally. E.g., in tests we should be able to create a dictionary of projects and users and pass that in to the IDP so that we get known responses when making requests to the IDP.

esi_leap/common/idp/__init__.py
larsks: Also, instead of explicitly testing for the value `True`, consider either (a) simply testing if the value is set:

```
if os.environ.get('ESI_DEBUG') is not None:
```

Or (b) be a little more flexible in what you accept as truthy values:

```
if os.environ.get('ESI_DEBUG', '').lower() in ['true', '1']:
```

esi_leap/common/idp/dummyIDP.py
QuanMPhm: @larsks Do you want me to add functions in `dummyIDP.py` or `baseIDP.py` to add projects or users?

QuanMPhm:
My apologies for the ugly diff. I should have waited to do linting and formatting at the end. 

I have re-written `dummyIDP.py` to now accept new projects and users at runtime. The option to pick an idp class has been added in `conf.esi.py`. The functions `offer_get_dict_with_added_info` and `lease_get_dict_with_added_info` has been moved per @larsks's suggestion.

The only question I have left is how the test cases should be modified. References to the  moved functions above will of course be changed, as well as references to the keystone idp, but I'd still like to ask...

- Will we be using `dummyIDP` or `keystoneIDP` in our test cases?
- Given the refactoring and the addition of the configurable idp option, should any new tests be added?

larsks:
> Will we be using dummyIDP or keystoneIDP in our test cases?

We will be using `dummyIDP` in the unit tests, because the point of unit tests is to remove external dependencies (we're not trying to test keystone itself) and verify the functionality of small units of code.

> Given the refactoring and the addition of the configurable idp option, should any new tests be added?

I'll have to think on that a bit.

esi_leap/common/idp/dummyIDP.py
```diff
@@ -0,0 +1,79 @@
...
+from keystoneclient.v3.users import UserManager
+from oslo_utils import uuidutils
+
+from esi_leap.common import exception
+from esi_leap.common.idp import baseIDP
+
+
+class DummyIDP(baseIDP.BaseIDP):
+
+    dummy_project_list: list[Project]
```
knikolla, line 26: I would make both of these dictionaries where the key is the `id`. This simplifies `get_project_name` and `_get_project_obj_from_id` significantly.

QuanMPhm:
https://github.com/CCI-MOC/esi-leap/blob/3288e635c976796520c1a16658063ea71f84c906/esi_leap/common/idp/__init__.py#L8-L11

@larsks I've made the function above to allow code in the esi_leap API to obtain the configured IDP class at runtime, instead of at import time, as we've talked about in our Tuesday meeting. Since test cases can [override](https://github.com/CCI-MOC/esi-leap/blob/3288e635c976796520c1a16658063ea71f84c906/esi_leap/tests/api/base.py#L33) the config file at runtime, this allows us to make the esi_leap API use arbitrary IDP classes.

https://github.com/CCI-MOC/esi-leap/blob/3288e635c976796520c1a16658063ea71f84c906/esi_leap/api/controllers/v1/utils.py#L143-L144

As shown above, with this change, code in the esi_leap API, such as the Controller classes, will obtain the idp by calling `get_idp()`. From our discussions, I believe you wanted the idp class to be passed to the Controller classes itself, or the entire API object, instead of having the idp obtained at the function level like above? If that is the case, I will do some more investigating to see if there's a simple way of doing so.

QuanMPhm:
@larsks In this draft update, I have done the following:
- Rebased the PR
- Made sure all test cases passed, which involved:
    - Changing every reference of `keystone` to `idp.dummyIDP.DummyIDP` in the test cases
    - [Overriding](https://github.com/CCI-MOC/esi-leap/blob/488dedd462d5b0549a276f99c24e7b4c9f84983b/esi_leap/tests/api/controllers/v1/test_lease.py#L634) `CONF.pecan.auth_enable` to True in `TestLeaseControllersGetAllFilters` (For some unknown reason, your commits passed this test class, but when I fetch from upstream and run the tests on my VM, upstream does not pass)
    - [Overriding](https://github.com/CCI-MOC/esi-leap/blob/488dedd462d5b0549a276f99c24e7b4c9f84983b/esi_leap/tests/api/controllers/v1/test_utils.py#L495-L496) the idp class to dummyIDP in `TestOfferLesseeUtils`

I will fix the python 3.8 test later, since it requires a very small fix.
With the tests (mostly) passed, I have two questions that I will wait on you before moving forward:
- I have just override the config options at two extra places, and had a mild feeling that these overrides could be better placed elsewhere. Should they be placed elsewhere?
- You mentioned in the past about refactoring the test cases. How should they be refactored? When I took a brief review of the test cases, many of them check if certain idp functions were called, in which case I don't see how we can avoid mocking the idp. I can see how it can be avoid for other cases though.

QuanMPhm:
@larsks Is this PR still being considered?

tzumainn:
I'm not positive about my comment here (and I'd be interested to see what others thing), but I think that the IDP abstraction here is a layer too high. There are only two keystone calls being made - `get` and `list` - and I think those are the functions that should be abstracted out. This allows a lot of the logic duplicated in the dummy and keystone IDP to be un-duplicated.

In my opinion, I think the right path would be to:

* change the IDPs to only abstract out project `get` and `list`
* rename `common/keystone.py` to something like `common/idp.py`
* have `common/idp.py` call the two IDP functions

larsks:
> Is this PR still being considered?

@QuanMPhm Yes, the past couple of weeks have just been a bit crazy. I'll try to take a look through the PR tomorrow.

QuanMPhm:
@tzumainn I abstracted the IDP instead of the individual Keystone functions based on @larsks' suggestion in the original issue (#159). Performing a text search over the `upstream/master` branch, it seems that every function in `common/keystone.py` was being called somewhere in `esi-leap`, which meant that all of those functions had to be abstracted away and implemented differently for `KeystoneIDP` and `DummyIDP`.

Thinking about your suggestion a bit more, I can sort of understand that the only function in `common/keystone.py` that is strongly coupled with the Keystone service is `get_keystone_client`, and I can see how it may also make sense to only abstract that function, instead of the entire `common/keystone.py`. @tzumainn @larsks I will follow whichever approach you guys prefer.

I have revised my PR to make it more complete and to make my vision of the final PR more clear. I've split the PR into 2 commits, since one only involves moving two functions around. I've also updated the commit message to better reflect the changes I added. Hopefully that will make this mess of a PR easier to follow.

tzumainn:
> @tzumainn I abstracted the IDP instead of the individual Keystone functions based on @larsks' suggestion in the original issue (#159). Performing a text search over the `upstream/master` branch, it seems that every function in `common/keystone.py` was being called somewhere in `esi-leap`, which meant that all of those functions had to be abstracted away and implemented differently for `KeystoneIDP` and `DummyIDP`.
> 
> Thinking about your suggestion a bit more, I can sort of understand that the only function in `common/keystone.py` that is strongly coupled with the Keystone service is `get_keystone_client`, and I can see how it may also make sense to only abstract that function, instead of the entire `common/keystone.py`. @tzumainn @larsks I will follow whichever approach you guys prefer.
> 
> I have revised my PR to make it more complete and to make my vision of the final PR more clear. I've split the PR into 2 commits, since one only involves moving two functions around. I've also updated the commit message to better reflect the changes I added. Hopefully that will make this mess of a PR easier to follow.

Every function in `common/keystone.py` is definitely being called somewhere in `esi-leap`, but that's only because we're the ones that created those functions.

It's important to recognize that `get_keystone_client` is a strong point of commonality, but by itself it's only the means to an end. The client by itself is not useful; the fact that we can call `client.projects.get` or `client.projects.list` is.

That's why I'd strongly recommend those two functions being the point of abstraction; they represent an actual API call to keystone, and they are fundamentally what need to be abstracted if we want to replace keystone with an arbitrary IDP. 

Thinking about it another way: right now your implementation abstracts out the functions in `common/keystone.py`, such as `get_parent_project_id_tree` and `get_project_name`. However much of those functions consist of ESI logic, which is now duplicated in each of your IDP implementations. If we instead abstract out the Keystone call that both of them use - `client.projects.get` - then we get rid of that duplication in logic.

QuanMPhm:
@tzumainn I have implemented your feedback, and will wait for further instructions before turning this draft into a PR.

esi_leap/common/idp.py
```diff
@@ -23,25 +22,26 @@
 _cached_project_list = None
 
 
-def get_keystone_client():
+def get_idp_client():
     global _cached_keystone_client
```
tzumainn, line 26: it probably makes sense to rename this to `_cached_idp_client`

esi_leap/common/idp.py
```diff
@@ -23,25 +22,26 @@
...
     if _cached_keystone_client is not None:
         return _cached_keystone_client
 
-    auth_plugin = ks_loading.load_auth_from_conf_options(CONF, "keystone")
-    sess = ks_loading.load_session_from_conf_options(CONF, "keystone", auth=auth_plugin)
-    cli = keystone_client.Client(session=sess)
+    # Get Client config option
+    module_path, class_name = CONF.esi.idp_plugin_class.rsplit(".", 1)
+    module = importlib.import_module(module_path)
+    cli = getattr(module, class_name)()
```
tzumainn, line 33: This feels a little delicate - can you do something similar to what we do for resource objects? (https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/objects/lease.py#L313 which calls https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/resource_objects/__init__.py#L34)

tzumainn (changes requested):
Hi! Just a few comments inline. In addition, `test_keystone.py` should probably be renamed `test_idp.py`, and you'll want unit tests for both the keystone and dummy implementation. The `resource_objects` do something similar.

QuanMPhm:
@larsks @tzumainn @knikolla What is the state of this PR? Is it not prioritized?

larsks:
@QuanMPhm I think the issue right now is that there's some uncertainty over the future of ESI in its current form. @tzumainn and I are have become heavily involved in an internal project for which one of the goals is making it easier to offer ESI-like services. I will try to take a closer look at your most recent changes here on Friday so that we can perhaps get this merged in.

PR #163, opened 2024-07-12
Add Additional Ironic Attributes to Nodes Resource

ajamias:
Resolves CCI-MOC/esi#571


larsks:
I like the commit message to contain (a) some explanation about why we're making these changes and (b) a reference of some sort to any associated issues. Something like:

```
Add target_provision_state and (target_)power_state attributes to nodes

The UI display of nodes currently calls both esi node list and Ironic
node list, merging the information together. The REST API
behind esi node list already grabs Ironic nodes and fetches some of
the needed information.

This commit modifies esi API to return power state information as part
of the node list operation, eliminating the redundant call to the
Ironic API.

Closes #163
```

ajamias:
Apologies, I just updated it

tzumainn (approved):
Looks good to me!

PR #164, opened 2024-07-12
Add pre-commit configuration and workflow

larsks:
Add a pre-commit [1] configuration in `.pre-commit-config.yaml` and a
corresponding GitHub workflow to run the checks on pull requests.

The pre-commit checks use ruff [2] for both linting and formatting.
Developers should integrate these checks into their local workflow by
installing pre-commit and running `pre-commit install` to enable the local
pre-commit hook.

Note that we have removed `tox -e pep8` from the existing test workflow, since
there's no reason to run the linting as part of the Python version matrix.

[1]: https://pre-commit.com/
[2]: https://docs.astral.sh/ruff/


larsks:
Sure, since this hasn't merged yet. I'll update the PR in a couple of minutes.

DanNiESh (approved):

tzumainn (commented):
* is " really *that* preferred over '??
* it feels like the README could use a note regarding checks that are run, and how to run them - is that something that could be added?

larsks:
> is " really that preferred over '??

Meh. The point of a formatter is to *pick one* and then use it so that all the code is consistent.

> it feels like the README could use a note regarding checks that are run, and how to run them 

Good thought; I will update the README.

tzumainn:
> > is " really that preferred over '??
> 
> Meh. The point of a formatter is to _pick one_ and then use it so that all the code is consistent. That is, it's not so much about preference as it is about consistentency. If you integrate the formatter in your editor you don't even have to think about it 😄.
> 
> > it feels like the README could use a note regarding checks that are run, and how to run them
> 
> Good thought; I will update the README.

Fair enough; I'll point out that Ironic seems to have picked the single quote over the double quote (and that's my random preference as well), but I don't really care enough to make you undo all the changes and push them the other way.

larsks:
@tzumainn I've added a section to the README.

tzumainn (commented):
Ah, I see how it works - neat! It seems pretty strict but whatever.

Should `pre-commit` be put in `test-requirements.txt`?

tzumainn:
> Ah, I see how it works - neat! It seems pretty strict but whatever.
> 
> Should `pre-commit` be put in `test-requirements.txt`?

And the answer is no, as you just explained to me! LGTM

tzumainn:
oh, except for the conflicts because I just merged the other thing

PR #166, opened 2024-07-17
Return Node's Future Leases and Future Offers as List

ajamias:
The future_leases attribute for a node is currently a concatenation of all future leases with no separation. By adding a newline, the user can more easily see and parse the future leases in the CLI and REST API. The future offers were previously separated by a space character, but changed to a newline for consistency.

### DIFF
--- a/esi_leap/api/controllers/v1/node.py
+++ b/esi_leap/api/controllers/v1/node.py
@@ -44,8 +44,8 @@ class Node(base.ESILEAPBase):
     offer_uuid = wsme.wsattr(wtypes.text)
     lease_uuid = wsme.wsattr(wtypes.text)
     lessee = wsme.wsattr(wtypes.text)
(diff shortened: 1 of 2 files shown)


tzumainn (commented):
It feels like `f_offer_uuids` and `f_lease_uuids` should maybe actually just be lists of related offers and leases; and the CLI client should just iterate through them and print whatever it wants. Basically, the API should return objects and leave the formatting of values up to the CLI (or UI).

tzumainn (commented):
I know this isn't your fault, but it is surprising to me that these changes don't affect the unit tests. I took a look and it looks like we never added checks to ensure the these offer and lease uuids were processed correctly. Can you add those?

After that, this looks good! I'll hold off on merging until after the corresponding CLI PR is merged though.

tzumainn (approved):
Looks good - thanks!

tzumainn (approved):
Looks good!

PR #167, opened 2024-07-19
Fix test warnings and errors under pytest

larsks:
This is prep work for adding some integration tests to this repository. I would like to be able to use pytest [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) as part of the integration tests, but in order to do that we need to ensure that the existing tests run cleanly under pytest.

I tried running the test suite using pytest, and while it mostly works, there were a number of errors and warnings that cropped up. Primarily:

1. We need to catch exceptions in threads spawned during the test process.

2. We need to pass valid UUIDs to anything expecting a UUID; failure to do so yields a warning. So many warnings.

With this pull request, all tests run cleanly under pytest. There are still warnings, but they all come from third party modules and so are somebody else's problem.

### DIFF
--- a/esi_leap/db/sqlalchemy/models.py
+++ b/esi_leap/db/sqlalchemy/models.py
@@ -14,7 +14,7 @@ from oslo_db.sqlalchemy import models
 from oslo_db.sqlalchemy import types as db_types
 
 from sqlalchemy.ext.compiler import compiles
-from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import declarative_base
 from sqlalchemy import orm
 from sqlalchemy import Column, DateTime, ForeignKey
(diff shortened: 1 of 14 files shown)


esi_leap/resource_objects/test_node.py
```diff
@@ -13,7 +13,7 @@
 from esi_leap.resource_objects import base
 
 
-class TestNode(base.ResourceObjectInterface):
+class FakeNode(base.ResourceObjectInterface):
```
tzumainn, line 16: Not a huge deal, but would it make sense to rename the file `fake_node.py`?

larsks: Sure, all done!

tzumainn (approved):
Looks good to me - thanks!

PR #168, opened 2024-07-19
Replace stestr with pytest

larsks:
Using pytest instead of stestr to run the tests gives us access to a broad range of test plugins and convenient testing features like pytest fixtures.

This commit includes the pytest-cov plugin for code coverage reporting; running the unit tests via `tox -e py3` will produce coverage statistics on stdout.


tzumainn (approved):
Looks good!

PR #169, opened 2024-07-22
Add integration tests for container image

larsks:
Verify that we can successfully build an esi-leap container image and that
we can successfully start a container from the image.

These tests are run automatically by .github/workflows/build-image.yaml,
which on pushes to the `master` branch will also push the image into the
ghcr.io container repository.

The integration tests introduced in this commit are minimal in nature and
are ripe for further work if someone is feeling inspired.

To run the integration tests manually:

    tox -e integration

Additionally, this PR ensures that github actions versions (a) are up-to-date and (b) are monitored by dependabot so that they will be kept current in the future.

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -1,0 +1,3 @@
+.tox
+.coverage
+.pytest_cache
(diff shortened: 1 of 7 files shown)


tzumainn (approved):
Thanks for doing this!

PR #171, opened 2024-07-23
Updated api-ref in esi-leap documentation

sheldor1510:
- [#566](https://github.com/CCI-MOC/esi/issues/566)

### DIFF
--- a/docs/esi-leap-api-ref.md
+++ b/docs/esi-leap-api-ref.md
@@ -2,27 +2,44 @@
 
 The Offer API endpoint can be reached at /v1/offers.
 
-##### GET
+##### GET /v1/offers/\<uuid_or_name> - Show Offer Details
 * The /v1/offers/\<uuid_or_name> endpoint is used to retrieve the offer with the given uuid. The response type is 'application/json'.
+
+
+##### GET /v1/offers - List Offers
 * The /v1/offers endpoint is used to retrieve a list of offers. This URL supports several URL variables for retrieving offers filtered by given values. The response type is 'application/json'.
   * project_id: Returns all offers with given project_id.
   * status: Returns all offers with given status.
     * This value will default to returning offers with status 'available'.
     * This value can be set to 'any' to return offers without filtering by status.
   * resource_uuid: Returns all offers with given resource_uuid.
-  * resource_type: Returns all offers with given resource_type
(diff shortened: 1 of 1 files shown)


docs/esi-leap-api-ref.md
```diff
@@ -10,58 +10,82 @@ The Offer API endpoint can be reached at /v1/offers.
     * This value will default to returning offers with status 'available'.
     * This value can be set to 'any' to return offers without filtering by status.
   * resource_uuid: Returns all offers with given resource_uuid.
-  * resource_type: Returns all offers with given resource_type
+  * resource_type: Returns all offers with given resource_type.
+  * resource_class: Returns all offers with given resource_class.
```
tzumainn, line 14: Might be worth mentioning here (and below) that this will default to the configured "default resource class", which by default is "ironic_node".

docs/esi-leap-api-ref.md
```diff
@@ -10,58 +10,82 @@ The Offer API endpoint can be reached at /v1/offers.
...
+            }
+          ]
         }
-      ]
-    }
-}'
-```
+    }'
+    ```
+* The /v1/offers/\<uuid>/claim endpoint supports POST requests to claim the offer with the given uuid. The body of the request contains the lease information with the given values.
```
tzumainn, line 78: I would format this section so there's an actual header, so it's more apparent that there's another API function here. Right now it's a bit lost within the `POST` section. You can take a look at https://docs.openstack.org/api-ref/baremetal/ for an example of how specialized API commands are listed.

tzumainn (changes requested):
Pretty good! Just a few minor comments.

docs/esi-leap-api-ref.md
sheldor1510: Mentioned.

docs/esi-leap-api-ref.md
sheldor1510: I added headers to each endpoint to make it structured in a similar way to the Baremetal API documentation. Let me know if this fine or too much.

tzumainn (changes requested):
Yep, that looks good! One last request: can you squash your commits together? We like doing that because it unclutters the commit history.

tzumainn (approved):
Looks good - thanks!

PR #176, opened 2024-10-08
Enforce default and max lease length times when offer is claimed

tzumainn:


tzumainn:
This is pretty much a bug that this wasn't enforced before.

DanNiESh (approved):
LGTM

PR #178, opened 2024-11-05
Clear instance_info when removing lease from a node

tzumainn:
If a node had instance_info set but was not deployed, then removing the lease would not clear the instance_info. This change fixes that issue.


skanthed:
looks good to me

tzumainn:
> looks good to me

Thanks! Can you approve by submitting a review?

PR #181, opened 2024-12-20
Update console auth token API to allow user to pass in token ttl

tzumainn:


tzumainn:
@DanNiESh could you also look at this one? I forgot to mention it's needed for the SDK.

PR #182, opened 2025-01-07
Update lease removal function to be more comprehensive

tzumainn:
This change updates the lease removal function to undeploy the node if it is in any deployed state, and to remove any vif attachments on the node otherwise.


DanNiESh (approved):
Looks good to me!

PR #186, opened 2025-03-26
Make dates timezone aware

tzumainn:
If one datetime is offset-aware and another is offset-naive, any comparision between the two will fail. This PR enforces updates these comparisions so that they are always offset-aware.

### DIFF
--- a/esi_leap/common/utils.py
+++ b/esi_leap/common/utils.py
@@ -10,6 +10,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+import datetime
 
 from oslo_concurrency import lockutils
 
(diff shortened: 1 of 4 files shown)


esi_leap/common/utils.py
```diff
@@ -19,3 +20,9 @@
 
 def get_resource_lock_name(resource_type, resource_uuid):
     return resource_type + "-" + resource_uuid
+
+
+def datetime_aware(dt):
+    if dt.tzinfo is not None and dt.tzinfo.utcoffset(dt) is not None:
```
larsks, line 26: When will `dt.tzinfo.utcoffset(dt)` return `None` when `dt.tzinfo` is not None?

larsks (approved):

esi_leap/common/utils.py
tzumainn: Not entirely sure myself, but the python documentation implies it can be:

https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive

